PR #7, opened 2021-12-25
Fix darwin silicon

gilligan:
- Fixes the build: darwin needs the Security framework
- Fixes the shell: pre-commit-hooks.nix is currently broken for apple silicon so this PR disables its use for this platform for now

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -2,6 +2,7 @@
 let
   pins = import ./npins;
   pkgs = import pins.nixpkgs { inherit system; };
+  inherit (pkgs) stdenv lib;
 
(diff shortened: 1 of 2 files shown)


shell.nix
```diff
@@ -31,6 +32,7 @@ pkgs.mkShell {
     nix-prefetch-git
     git
   ];
-
+} // lib.optionalAttrs (stdenv.isDarwin && stdenv.isAarch64) {
```
andir, line 35: Isn't this missing a `!` ?

gilligan: yup, it most certainly is

PR #11, opened 2021-12-26
Add versioning support

### LINKED ISSUE #10: Bikeshedding
I suggest renaming the `pins.json` file to `sources.json`. I personally find it more pleasing, but it would also be more consistent with Niv (and what I'd like to see in nixpkgs later on).

Furthermore, I propose renaming all `hash` fields in the JSON file to `XXX_hash`, in order to make more clear what the hash is about. So for git this will be `revision_hash` and for GitHub this will be `tarball_hash`. At the moment, there is some (slight) conceptual ambiguity because git revisions are also hashes. (For example, the `hash` field for GitHub pins could either be the hash of the tarball or the git commit.)

piegamesde:


piegamesde:
The other changes are unrelated and got added more or less by accident. I can include them if you like to, otherwise I'll drop them (for now, of course).

readme
```diff
@@ -0,0 +1 @@
+/nix/store/7wsnrffjppv7yh81ams8gx6z3zb8l3mz-readme
```
andir, line 1: should not be committed!

PR #15, opened 2021-12-27
Update readme

gilligan:
Provide a more fleshed out README

- Provide some badges
- Provide a usage section
- Change the command reference to show the command instead of `help <command>`


gilligan:
@andir The `license` badge will report that no license is detected until we pick one ;-)

piegamesde:
Please rebase on latest master, I think the CLI options have changed a bit since then.

Also, how about adding a badge for our Matrix room (#npins:kack.it)?

PR #30, opened 2022-05-20
Nixpkgs mode

piegamesde:
For usage within nixpkgs, we want the `sources.json` to be level with the current folder, and we don't need a `default.nix` since it will be included globally in nixpkgs itself.


piegamesde:
Implemented in #120

PR #33, opened 2022-08-10
Add support for gitlab private repositories

dasJ:
We kind of need this

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,9 +50,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
+checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "base64"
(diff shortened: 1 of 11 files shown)


piegamesde:
Thank you, this looks pretty solid. Some quite notes:

- If the token has other uses other than private repositories (e.g. rate limiting), maybe a more generic name like `auth_token` or `access_token` would be preferable?
- How does the Nix side fetcher handle this? That change is missing from the PR
- I'll have to think about that some more, but it's pretty likely that this will require incrementing the format version again.

dasJ:
- I actually had the field named `access_token` during development but I changed it to `private_token`. This is [the relevant docs](https://docs.gitlab.com/ee/api/#personalprojectgroup-access-tokens) so if you want me to change it, I will
- The fetcher handles this as the `?private_token=…` is appended to the URL
- I **think** the format should be backwards compatible as older npins versions just ignore the field

piegamesde:
Since the added field may be required for private repositories to successfully fetch, older versions can't really ignore it. Therefore, I'm in favor of incrementing the format version.

andir:
Please don't forget to add unit tests for the new fields/logic in the GitLab case. We shouldn't let those fall behind. Especially if we can't (easily) have an integration tests those should be present.

dasJ:
Added tests, added more docs, added token via environment, bumped the version

andir (approved):

dasJ:
Andi mach release

piegamesde:
piegames macht release

PR #34, opened 2022-10-22
Add a plugin for Hydra

dasJ:


contrib/hydra/NpinsInput.pm
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    # Parse JSON
+    my $json_contents;
+    if (not eval { $json_contents = decode_json(read_text($sources_json)); 1 }) {
+        _print($evaluation, 'Invalid JSON');
+        return;
+    }
+
+    if ($json_contents->{version} != 2) {
```
andir, line 97: Unfortunately, this isn't compatible with the current version of npins: https://github.com/andir/npins/commit/fe8e6fe06c17529de7dcf67c71ca43c7adc6a50c ;)

andir:
As discussed yesterday, we might want to have this in hydra instead? In any case, there must be tests before this can be added.

andir:
@dasJ any movement on this?

PR #36, opened 2022-11-25
Fix private repos and add filename support

dasJ:
I'm very sorry for this.

So essentially when using a private repos, the `sha1=` parameter was **replaced** by the private token rather than the private token being appended to the query params.
This would have been an easy fix but that actually results in `&` being part of the derivation name which is invalid. This is why I now generate filenames for all the pins that are fetchurl'ed. Yay.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -24,6 +24,7 @@ impl Pin {
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct ChannelVersion {
     pub url: url::Url,
+    pub filename: String,
 }
 
 impl diff::Diff for ChannelVersion {
@@ -62,15 +63,35 @@ impl Updatable for Pin {
             .url()
             .clone();
 
-        Ok(ChannelVersion { url })
+        Ok(ChannelVersion {
+            url: url.clone(),
+            filename: Self::calc_filename(&url)?,
+        })
     }
 
     async fn fetch(&self, version: &ChannelVersion) -> Result<ChannelHash> {
         /* Prefetch an URL that looks like
          * https://releases.nixos.org/nixos/21.11/nixos-21.11.335807.df4f1f7cc3f
          */
-        let hash = nix::nix_prefetch_tarball(&version.url).await?;
+        let hash =
+            nix::nix_prefetch_tarball(&version.url, Some(Self::calc_filename(&version.url)?)).await?;
 
         Ok(ChannelHash { hash })
+    }
+}
+
+impl Pin {
+    pub fn calc_filename(url: &url::Url) -> Result<String> {
+        let name_parts = url.path_segments().unwrap().collect::<Vec<&str>>();
+        let basename = name_parts
+            .get(1)
+            .ok_or_else(|| anyhow::format_err!("Unsupported channel URL {}", url))?;
+        let ending = name_parts
+            .last()
+            .ok_or_else(|| anyhow::format_err!("Unsupported channel URL {}", url))?
+            .split('.')
+            .skip(1)
+            .collect::<Vec<&str>>();
+        Ok(format!("npins-channel-{}.{}", basename, ending.join(".")))
     }
 }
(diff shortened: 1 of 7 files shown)


src/git.rs
```diff
@@ -203,14 +207,25 @@ impl Repository {
...
         })
     }
+
+    fn filename(&self) -> Option<String> {
+        match self {
+            Repository::GitLab { repo_path, .. } => Some(format!(
+                "npins-gitlab-{}.tar.gz",
+                repo_path.split('/').last().unwrap().to_owned()
+            )),
+            Repository::GitHub { repo, .. } => Some(format!("npins-gitlab-{}.tar.gz", repo)),
```
Conni2461, line 225: should be `"npins-github-{}.tar.gz"`

src/git.rs
```diff
@@ -284,12 +299,17 @@ impl Updatable for GitPin {
 
     async fn fetch(&self, version: &GitRevision) -> Result<OptionalUrlHashes> {
         let url = self.repository.url(&version.revision)?;
+        let filename = self.repository.filename();
         let hash = match url.as_ref() {
-            Some(url) => nix::nix_prefetch_tarball(url).await?,
+            Some(url) => nix::nix_prefetch_tarball(url, filename.clone()).await?,
```
Conni2461, line 304: you don't need to clone here you can do `filename.as_ref()`

src/git.rs
```diff
@@ -203,14 +207,25 @@ impl Repository {
...
                 if let Some(token) = private_token {
-                    url.set_query(Some(&format!("private_token={}", token)));
+                    url.query_pairs_mut().append_pair("private_token", token);
                 }
                 Some(url)
             },
         })
     }
+
+    fn filename(&self) -> Option<String> {
```
andir, line 219: Tests pretty please.

src/git.rs
```diff
@@ -203,14 +207,25 @@ impl Repository {
...
+        match self {
+            Repository::GitLab { repo_path, .. } => Some(format!(
+                "npins-gitlab-{}.tar.gz",
+                repo_path.split('/').last().unwrap().to_owned()
```
andir, line 223: What are our assurances that we can really unwrap here? Do we verify this while unserializing? Do we verify it (only?) during interactive modification of the JSON file?

src/channel.rs
```diff
@@ -62,15 +63,36 @@ impl Updatable for Pin {
...
+        let hash =
+            nix::nix_prefetch_tarball(&version.url, Some(Self::calc_filename(&version.url)?))
+                .await?;
 
         Ok(ChannelHash { hash })
     }
 }
+
+impl Pin {
+    pub fn calc_filename(url: &url::Url) -> Result<String> {
```
andir, line 85: Tests please.

src/channel.rs
```diff
@@ -62,15 +63,36 @@ impl Updatable for Pin {
...
+        let name_parts = url.path_segments().unwrap().collect::<Vec<&str>>();
```
andir, line 86: What makes us think it is safe to unwrap here?

andir (changes requested):
Just a quick pass. We need some test coverage and commentary around the new code.

piegamesde:
I really dislike having to track the name in the pins JSON. Is there no way that we can solve this in Nix on the import side of things with a simple string substitution?

dasJ:
I can try but I don't see how this will be fast

dasJ:
Ah yes we have to do this in the Rust code because `nix-prefetch-url` will otherwise fail because the path is invalid

piegamesde:
At this point the support for private repos has bitrotten quite a bit more so I'm considering removing it again altogether

samueldr:
(Doing some triage)

I believe the filename thing is helped/solved/worked around since b131c384d06ac80307a0628148174dc30c3d45d4

Which anyway as documented since that commit, ensures that the Nix `builtin` doesn't end-up re-fetching the source.

So assuming that the private repo support is either removed or to be removed (as per previous comment), I think that leaves nothing to be rebased from this PR, right?

PR #38, opened 2022-11-25
Darwin fixes

andir:


shell.nix
```diff
@@ -32,11 +32,10 @@ pkgs.mkShell
     nix_2_3
     nix-prefetch-git
     git
-  ];
+  ] ++ (if stdenv.isDarwin then [
```
gilligan, line 35: How about using `lib.optional` here instead?

gilligan (approved):
minor comment but looking good

Issue #37, opened 2022-11-25
idea: fetch build system specific dependency metadata


andir:
For Rust and Go (and other languages) it is common to either have a big IFD derivation as part of the nix build or to have a `2nix` tool generate a ton of (garbage) code. Especially in nixpkgs or pure-eval mode that is a big deal as you can't read the lock files within the repositories that you are trying to package.

For at least Rust (and others, with some effort) we could add additional source information in the lockfile.

I'll try to illustrate this with a Rust project called `ripgrep`.

When you run `npins add github burntsushi ripgrep` `npins` will track the latest release of the repository, fetch tarball and store the hash in the lockfile. When you package `ripgrep` with Nix you need to provide the crates `ripgrep` depends on. As describe above that is usually IFD or some convoluted code generation step. What `npins` could be doing in this case is parsing the `Cargo.lock` toml and generate additional metadata as part of the lockfile.

We should add a `--extra-metadata rust,npm,....` switch to the add command.

This would be the same that `import-cargo` (https://github.com/edolstra/import-cargo/blob/25d40be4a73d40a2572e0cc233b83253554f06c5/flake.nix#L8-L68) or what `buildRustPackage` with `cargoLock.lockFile = ...` in Nixpkgs (https://github.com/NixOS/nixpkgs/blob/ce1f9354959ae1493916f2e551ecc32e79b4a473/pkgs/build-support/rust/import-cargo-lock.nix) do when they can access the Cargo.lock file.

In the lockfile it could look like this (pseudo code):
```yaml
{
   "ripgrep": {
       "type": "github",
       /* common attributes */
      "extraAttributes": {
         "cargoDependencies": {
             "atty-0.2.14": /* we could drop the name and return a list, this example name matches with nixpkgs would pick right now */ {
                /* all the values "raw" as they are in the Cargo.lock, we don't want to remove data or mangle it */
                "name": "atty",
                "version": "0.2.14",
                "source": "registry+https://github.com/rust-lang/crates.io-index",
                "checksum": "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8",
                "dependencies":  ["hermit-abi", "libc", "winapi", ]
         }
     }
   }
}
```

In our Nix code that parses the lockfile we could then attach those as additional (passthru-style) parameters to the output path just like we already provide `version`, `revision` etc...

This should be rather simple to add as we are already downloading the sources. We then have to read the Cargo.lock file (in the root of the repository, for now) and read the metadata.

PR #40, opened 2022-11-26
HACK: introduce metadata for newly added pins

andir:
HACK: introduce metadata for newly added pins

This is an experiment that includes Cargo.lock metadata in our
lockfiles for all added pins. The idea is outlined in [an issue] and
this is just a quick hack to get this going.

Currently the GitLab tests are failing as I didn't bother fixing them
given that they are currently being changed.

To test this initialize a new bare npins folder:

$ npins -d /tmp/foo init --bar

Then you can add a rust project (e.g. npins):

$ npins -d /tmp/foo add github andir npins

Now you should have a regular pin entry but also the JSON equivalent
of the Cargo.lock (TOML) file in a sub attribute.

$ less /tmp/foo/sources.json

This in itself isn't particular helpful yet but with a bit more
work (on our Nix shim and a variant of import-cargo) we can implement
the vision of the linked issue experimentally.

Open tasks:

 * [ ] Is there a better way to handle the prefetched source path?
 * [ ] Add some CLI options to opting into some kind of metadata
       inclusion, right now we just include it unconditionally.
 * [ ] Ergonomics, ergonomisc and ergonomics of the feature. It must
       not suck!

### DIFF
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -1,17 +1,11 @@
 use anyhow::{Context, Result};
 
-#[allow(unused)]
 pub struct PrefetchInfo {
-    store_path: String,
-    hash: String,
+    pub store_path: String,
+    pub hash: String,
 }
 
-pub struct NixPrefetchTarballResult {
-    pub sha256: String,
-    pub path: String,
-}
-
-pub async fn nix_prefetch_tarball(url: impl AsRef<str>) -> Result<NixPrefetchTarballResult> {
+pub async fn nix_prefetch_tarball(url: impl AsRef<str>) -> Result<PrefetchInfo> {
     let url = url.as_ref();
     let output = tokio::process::Command::new("nix-prefetch-url")
         .arg("--unpack") // force calculation of the unpacked NAR hash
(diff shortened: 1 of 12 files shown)


src/nix.rs
```diff
@@ -30,7 +30,16 @@ pub async fn nix_prefetch_tarball(url: impl AsRef<str>) -> Result<String> {
     Ok(String::from(stdout.trim()))
 }
 
-pub async fn nix_prefetch_git(url: impl AsRef<str>, git_ref: impl AsRef<str>) -> Result<String> {
+#[derive(Debug)]
+pub struct NixPrefetchGitResponse {
+    pub sha256: String,
+    pub path: String,
```
gilligan: How about using `storePath` instead of just `path`, for clarity?

src/nix.rs
```diff
@@ -6,10 +6,16 @@ pub struct PrefetchInfo {
     hash: String,
 }
 
-pub async fn nix_prefetch_tarball(url: impl AsRef<str>) -> Result<String> {
+pub struct NixPrefetchTarballResult {
+    pub sha256: String,
+    pub path: String,
```
gilligan: Same suggestion as before: maybe use `storePath` ?

samueldr:
> The idea is outlined in [an issue]
> 
> [...]
> 
> of the linked issue

:thinking:

andir:
> > The idea is outlined in [an issue]
> > [...]
> > of the linked issue
> 
> 🤔

Good point. I think it might be this: https://github.com/andir/npins/issues/37

auscyber:
what happened to this?

andir:
It mostly never left my PoC stage.. I've not had a usage for this but perhaps I'll revisit that later.

PR #48, opened 2023-08-03
Allow initializing pins without creating the default.nix

andir:
This allows usage of npins for locking-only purposes where the consumer might have a copy of a default.nix already or where they provide their own implementation (and will fix breakage caused by newer versions of npins).

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,8 +146,9 @@ USAGE:
     npins init [FLAGS] [OPTIONS]
 
 FLAGS:
-        --bare    Don't add an initial `nixpkgs` entry
-    -h, --help    Prints help information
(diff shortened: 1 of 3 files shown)


andir:
This is similar to #30 but a bit more light weight. We can later on add another exectuable that works better for the use cases that #30 was aiming for.

piegamesde:
How does this interact with the upgrade command?

bew:
For once this PR is merged, something that I wanted to have for a while is to be able to not even create a folder, and only have a single file (with configurable name, like how the dir is currently configurable) in working directory 🤔

piegamesde:
Implemented in #120

Issue #57, opened 2023-12-17
Adding a source does not seem to add it to the store


fricklerhandwerk:
When I add a large repo as a source, it has to download for a long time. Using that source in a derivation requires downloading it again, apparently. Is that a user error or is it how the tool works?

piegamesde:
I think that we try to use the correct Nix prefetch call to avoid this, but it might depend on the exact situation. Can you please show a command which adds a source which reproduces this?

eclairevoyant:
I've run into a similar issue myself with the following pin:

```json
    "google-fonts": {
      "type": "Git",
      "repository": {
        "type": "GitHub",
        "owner": "google",
        "repo": "fonts"
      },
      "branch": "main",
      "revision": "3992c83a1afd00abb72a6c237159f20e419433dc",
      "url": "https://github.com/google/fonts/archive/3992c83a1afd00abb72a6c237159f20e419433dc.tar.gz",
      "hash": "1zcqm82c7a08mkkfhm7py494qa41q8pb7x237hdqas01qw5kzcp8"
    },
```

Updating the pin seems to take a while on its own - presumably the tarball must be downloaded to get the hash.
First build after updating the pin also takes a while, as it tries to redownload the tarball.
The tarball itself is massive, so the time spent is noticeable.

fricklerhandwerk:
Well of course: `npins` [relies on `nix-prefetch-url`](https://github.com/andir/npins/blob/1e1495a5beeeb328f26dc1c34077969ff42e1546/src/nix.rs#L9-L18) and that in turn doesn't cache anything. When `builtins.fetchTarball` and such are called on the pinned URL, the whole download has to start over.

fricklerhandwerk:
Could it be that this is because `npins` relies on `nix-prefetch-url`, which doesn't use the fetcher cache?

piegamesde:
Yes indeed we shell out to the `nix-prefetch-url` and `nix-prefetch-git` commands (<https://github.com/andir/npins/blob/master/src/nix.rs>). I am a bit confused that these do not cache the result in the store. Are there better alternatives?

fricklerhandwerk:
`nix-prefetch-url` produces a store object but doesn't use the fetcher cache, which means that `builtins.fetchurl` will always get a cache miss and only find that the resulting store object already exists after the source was downloaded.

piegamesde:
Sounds like a nix bug to me tbh

fricklerhandwerk:
[...]

PR #74, opened 2024-05-06
builtins.fetchGit: use same name as nix-prefetch-git

stuebinm:
When using npins with (generic, i.e. non-github) git repositories, we currently have the issue of having to download these twice: once while doing `npins add git ...`, and then a second time when evaluating the source attribute in Nix.

The source of this is that `nix-prefetch-git`, which we use to get the hash, sets as name for the store path it produces a name derived from the url; this [behaviour is not configurable](https://github.com/NixOS/nixpkgs/blob/c215fb18dc1c4ef927c773078d2365fc559365c5/pkgs/build-support/fetchgit/nix-prefetch-git#L142). In contrast, `builtins.fetchGit` sets the name as `"source"` by default.

The nixpkgs fetcher uses [a nix implementation of the same logic](https://github.com/NixOS/nixpkgs/blob/2116a1123122d639b7c69e47284e4b35198c3fa6/pkgs/build-support/fetchgit/default.nix#L2) to derive the desired name in nix.

This commit adds a function which does the same, derived from the one in Nixpkgs, but tweaked to only use builtin functions, and uses it to set an appropriate name for the git fetcher. With it, a the name exposed by npins to nix code changes to the one that `nix-prefetch-git` also used, thus we avoid this doubled download after adding a new source.

This logic is somewhat delicate. I've added a safe guard so that if it ever fails to extract a name, we fall back to using `"source"` as attribute name.

Note that while this is a similar issue to https://github.com/andir/npins/issues/57, it is (afaict, anyways) not related to `builtins.fetchGit` not using the fetcher cache. That's a separate issue which also causes doubled downloads.

### DIFF
--- a/npins/default.nix
+++ b/npins/default.nix
@@ -27,6 +27,7 @@ let
       revision,
       url ? null,
       hash,
+      branch ? null,
       ...
     }:
     assert repository ? type;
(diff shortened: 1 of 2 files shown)


andir (commented):
Thanks for the fix! This sounds really useful! I'm currently curious how to keep this from regressing going forward. Could you come up with one or two tests that assure that the output names are as expected?

stuebinm:
I've added some tests, which should hopefully cover all uses of `npins add git` which actually occur. Incidentally, they already helped me find a bug in my initial commit, so seems like having them is definitely worthwhile :)

PR #78, opened 2024-05-11
Implement freezing and thawing of pins

### LINKED ISSUE #62: Pinning specific commits in a GH repo
To add a GH source at a specific commit is explained as

```sh
npins add --name <name> github <owner> <repo> -b <branch> --at <rev>
```

In my case, I'd like to _keep_ that source pinned to `<rev>` even after running `npins update`. Is this possible?

andir:
This allows freezing a pin to ensure that it isn't changed by issuing a blanked `npins update`.

Should fix #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ OPTIONS:
 
 SUBCOMMANDS:
     add             Adds a new pin entry
+    freeze          Freeze a pin entry
     help            Prints this message or the help of the given subcommand(s)
     import-flake    Try to import entries from flake.lock
     import-niv      Try to import entries from Niv
@@ -122,6 +123,7 @@ SUBCOMMANDS:
                     `default.nix` and never touch your sources.json
     remove          Removes one pin entry
     show            Lists the current pin entries
+    thaw            Thaw a pin entry
     update          Updates all or the given pin to the latest version
     upgrade         Upgrade the sources.json and default.nix to the latest format version. This may occasionally
                     break Nix evaluation!
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -387,6 +387,13 @@ pub struct ImportFlakeOpts {
 }
 
 #[derive(Debug, StructOpt)]
+pub struct FreezeOpts {
+    /// Names of the pin(s)
+    #[structopt(required = true)]
+    pub names: Vec<String>,
+}
+
+#[derive(Debug, StructOpt)]
 pub enum Command {
     /// Intializes the npins directory. Running this multiple times will restore/upgrade the
     /// `default.nix` and never touch your sources.json.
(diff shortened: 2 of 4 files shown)


eclairevoyant (approved):
Seems to work as expected. Thanks for looking into this!

README.md
```diff
@@ -115,13 +115,15 @@ OPTIONS:
...
     add             Adds a new pin entry
+    freeze          Freeze a pin entry
     help            Prints this message or the help of the given subcommand(s)
     import-flake    Try to import entries from flake.lock
     import-niv      Try to import entries from Niv
     init            Intializes the npins directory. Running this multiple times will restore/upgrade the
                     `default.nix` and never touch your sources.json
     remove          Removes one pin entry
     show            Lists the current pin entries
+    thaw            Thaw a pin entry
```
gilligan, line 126: I get that logically `freeze` <-> `thaw` makes sense but from a UX perspective I would prefer to have a more explicit indication that those two commands belong together.

Hence I'd prefer `freeze` and `unfreeze` - WDYT @andir  ?

piegamesde: I'd also prefer `unfreeze`, with an optional alias to `thaw` for those who find that more intuitive

piegamesde (dismissed):
Looking good, only some nits (and "thaw" as name)

piegamesde:
Now that I actually think of it, what is the workflow of manually updating frozen pins? Unfreezing them just for that purpose seems a bit unwieldy.

src/cli.rs
```diff
@@ -589,6 +602,11 @@ impl Opts {
 
         if opts.names.is_empty() {
             for (name, pin) in pins.pins.iter_mut() {
+                if pin.is_frozen() {
+                    log::info!("Skipping pin {} as it is frozen.", name);
```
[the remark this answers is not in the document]
andir: Yeah, I was planning on doing this for the entire code base. We've a few cases where that isn't the case.

andir:
> Now that I actually think of it, what is the workflow of manually updating frozen pins? Unfreezing them just for that purpose seems a bit unwieldy.

Perhaps an extra argument to the `update` command?

src/cli.rs
```diff
@@ -691,9 +717,13 @@ impl Opts {
...
             let (mut style, status) = if opts.names.is_empty() || valid_names.contains(name) {
-                (ContentStyle::new().grey(), "queued")
+                if pin.is_frozen() && !opts.update_frozen {
+                    (ContentStyle::new().dark_grey(), "frozen")
+                } else {
+                    (ContentStyle::new().grey(), "queued")
+                }
             } else {
                 (ContentStyle::new().dark_grey(), "ignored")
             };
```
Coca162, line 729: Would probably nicer turned into a match now:
```rust
let (mut style, status) = match (
    opts.names.is_empty() || valid_names.contains(name),
    pin.is_frozen() && !opts.update_frozen,
) {
    (true, false) => (ContentStyle::new().grey(), "queued"),
    (true, true) => (ContentStyle::new().dark_grey(), "frozen"),
    (false, _) => (ContentStyle::new().dark_grey(), "ignored"),
};
```

Coca162 (approved):
Doesn't look like it breaks anything on the updating side :+1:

PR #80, opened 2024-06-02
fix build on x86_64-darwin

fricklerhandwerk:
also "works on my machine" now

### DIFF
--- a/npins.nix
+++ b/npins.nix
@@ -57,7 +57,13 @@ let
 
     inherit src;
 
-    buildInputs = lib.optional stdenv.isDarwin (with darwin.apple_sdk.frameworks; [ Security ]);
+    buildInputs = lib.optional stdenv.isDarwin (
+      with darwin.apple_sdk.frameworks;
+      [
+        Security
+        SystemConfiguration
+      ]
+    );
     nativeBuildInputs = [ makeWrapper ];
 
     # (Almost) all tests require internet


andir:
Should we add a Darwin build to CI? Does the free GHA plan allow this?

fricklerhandwerk:
It seems they do: https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories

phanirithvij:
@fricklerhandwerk does `nix-build -A meta.tests` work on your mac? It is failing in gha, I don't have a mac just used the gha to debug.

fricklerhandwerk:
@phanirithvij indeed, the network mocks seem to be failing, but at first glance that looks more like a race condition. Specifically, there's

```log
OSError: [Errno 48] Address already in use
Connection to 127.0.0.1 8000 port [tcp/irdmi] succeeded!
```

... and then the git fetcher seems to get 404 on some items.

phanirithvij:
I did some basic changes and here's the diff https://github.com/andir/npins/compare/master...phanirithvij:npins:drawin-ci-temp and the test is hanging https://github.com/phanirithvij/npins/actions/runs/9439870618/job/25998498494, not sure if it helps.

<details>
<summary>not helpful</summary>
I have never owned/used a mac or plan to, I was just curious if I could get it working.
</details>

andir:
> @phanirithvij indeed, the network mocks seem to be failing, but at first glance that looks more like a race condition. Specifically, there's
> 
> ```
> OSError: [Errno 48] Address already in use
> Connection to 127.0.0.1 8000 port [tcp/irdmi] succeeded!
> ```
> 
> ... and then the git fetcher seems to get 404 on some items.

Perhaps worth considering a random port for these kinds of tests. That might already fix it? not sure what might be running on random macos box without a sandbox.

phanirithvij:
I changed the all the mkGitTest invocations to use a different port 9000 and the githubtest ones still use 8000. This change let the tests run but they didn't exit properly and I had to stop the workflow run after 30+ mins. (Not getting 404 anymore) https://github.com/andir/npins/pull/80#issuecomment-2157819821

samueldr:
This can be closed:

 - b6f34245408412ec7a3ef022d583a08ed937868d

PR #81, opened 2024-06-02
make npins directory configurable in the package recipe

fricklerhandwerk:
this is convenient if you want to manage sources in a custom directory,
and have npins in the shell environment for all contributors to use:

```nix
{ pkgs }:
let
  custom-npins = pkgs.npins.override { directory = "nix"; };
in
pkgs.mkShell { packages = [ custom-npins ]; }

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,8 +110,9 @@ FLAGS:
     -V, --version    Prints version information
 
 OPTIONS:
-    -d, --directory <folder>    Base folder for sources.json and the boilerplate default.nix [env: NPINS_DIRECTORY=]
-                                [default: npins]
+    -d, --directory <folder>    Base folder for sources.json and the boilerplate default.nix The default value can also
+                                be configured in the package recipe: npins.override { directory = "some-directory"; }
+                                [env: NPINS_DIRECTORY=]  [default: npins]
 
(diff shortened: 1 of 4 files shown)


piegamesde:
I agree with making this an environment variable, but why at build time though? (Also nit but IMO this is not build time, but packaging time or something, as the actual binary is not affected by it)

fricklerhandwerk:
Fixed the wording in the commit. The directory is already configurable through an environment variable. This change merely sets a customisable default value in the package recipe so the thing can be used in a shell without going lenghts as I did here, just to tell npins to work in `./nix`:

- https://github.com/fricklerhandwerk/lazy-drv/blob/main/default.nix#L20-L28
- https://github.com/fricklerhandwerk/lazy-drv/blob/main/nix/npins.nix

piegamesde:
hm then I don't fully understand your use case, why couldn't you just export an env variable in the shell hook?

fricklerhandwerk:
🤦 Of course.

```nix
{ pkgs }:
pkgs.mkShell {
  packages = with pkgs; [
    npins
  ];
  env = {
    NPINS_DIRECTORY = "nix";
  };
}

PR #91, opened 2024-07-05
shell: add cargo-expand to help see what the Rust macros do

philiptaron:
To try it out, run `cargo expand --color=always | less -R` and page through the output.

In particular, I wanted to see what the compiler saw with the `mkPins` macro.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -42,6 +42,7 @@ pkgs.mkShell {
     with pkgs;
     [
       cargo
+      cargo-expand
       clippy
       rustc
       rust-analyzer


andir:
Whats the answer you got from your questions? Should we add more documentation instead?

philiptaron:
It's been enough time that whatever question I had has fallen out of memory. In general, I find documentation only so good -- and seeing what the compiler or machine sees, [Godbolt-style](https://godbolt.org/), to be more instructive. This PR fits in that genre.

PR #95, opened 2024-07-21
Add support for Forgejo

crop2000:


piegamesde:
Thank you, that looks like a pretty good start already. How does forgejo differ from Gitea? Could we maybe even support both simply by adding an alias to the CLI subcommand?

Also, please add some tests, both a Rust test (I'd suggest using the Lix repo as test subject) and a mock test in `test.nix`

crop2000:
i did hand test it with lix, and can write some Rust test and a mock tests.
I don't know about the differences to gittea. 
But as what we are using is fairly similar to github i don't expect that they diverged on this topic.

crop2000:
I tend to feel that adding gitea support via a alias is just cluttering the cli ux ... so i didn't add an alias.

src/cli.rs
```diff
@@ -140,6 +140,61 @@ impl GitHubAddOpts {
...
+}
+impl ForgejoAddOpts {
+    pub fn add(&self) -> Result<(String, Pin)> {
+        let server_url = match Url::parse(&self.server) {
+            Ok(url) => url,
+            Err(firsterror) => match Url::parse(&("https://".to_string() + self.server.as_str())) {
+                Ok(url) => url,
+                Err(_seconderror) => return Err(firsterror.into()),
+            },
+        };
```
piegamesde, line 160:
```suggestion
        let server_url = Url::parse(&self.server)
            .or_else(|err|
                Url::parse(&("https://".to_string() + self.server.as_str())
                    .map_err(|_| err)
            )
```

maybe something more like this is more concise?

src/git.rs
```diff
@@ -143,6 +153,20 @@ impl Repository {
...
+                server,
+                owner,
+                repo,
+            } => Some(
+                format!(
+                    "{server}{owner}/{repo}/archive/{revision}.tar.gz",
+                    server = server,
+                    owner = owner,
+                    repo = repo,
+                    revision = revision,
```
piegamesde, line 166:
```suggestion
                    "{server}{owner}/{repo}/archive/{revision}.tar.gz",
```

Those shouldn't be necessary since a couple of Rust versions

src/git.rs
```diff
@@ -185,6 +209,20 @@ impl Repository {
...
+                server,
+                owner,
+                repo,
+            } => Some(
+                format!(
+                    "{server}api/v1/repos/{owner}/{repo}/archive/{tag}.tar.gz",
+                    server = server,
+                    owner = owner,
+                    repo = repo,
+                    tag = tag,
```
piegamesde, line 222:
```suggestion
                    "{server}api/v1/repos/{owner}/{repo}/archive/{tag}.tar.gz",
```

test.nix
```diff
@@ -132,6 +132,78 @@ let
...
+          jq
+        ];
+      }
+      ''
+        set -euo pipefail
+        export HOME=$TMPDIR
+        export NIX_STATE_DIR=$TMPDIR
+        export NIX_DATA_DIR=$TMPDIR
+        export NIX_STORE_DIR=$TMPDIR
+        export NIX_LOG_DIR=$TMPDIR
```
piegamesde, line 162: Maybe move these into the prelude to reduce code duplication?

piegamesde (approved):
Looking mostly good to me, just some code nits

crop2000:
are you interested in this pr?

piegamesde (changes requested):
Yes, still interested. Just spread a bit thin at the moment unfortunately

piegamesde (approved):

crop2000:
:)

PR #98, opened 2024-08-07
git: fix fetch on branch

thillux:
npins was not able to use branches different from the default branch in my use case. A test is added, which tries to use test-branch, while the default upstream branch is main. I had to also fix the default branch for other tests while doing this.

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -73,7 +73,7 @@ let
       url ? null,
       submodules,
       hash,
-      branch ? null,
+      branch ? "HEAD",
       ...
     }:
     assert repository ? type;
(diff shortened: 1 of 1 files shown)


thillux:
Do you have any idea, whats failing in the `forgejoSubmodule` test? I can't reproduce this locally.

piegamesde:
With the changes you've made, HEAD of the generated git repo points to a different rev than previously. You need to either restore the previous HEAD position, or update all test cases to point to a different commit. (Disclaimer: I only briefly looked at the diff and test log, I might be wrong)

piegamesde:
Please rebase again, I fixed a bug which was causing weird test failures

thillux:
I haven't managed to fix the tests and no longer use npins. Closing therefore.

piegamesde:
Thank you for your efforts.

PR #99, opened 2024-08-09
add local development override option similar to niv

thillux:


andir:
Could you write some documentation (at least in the README.md) and add a test for this?

thillux:
Sure (README.md) & I'll try (test).

thillux:
> Could you write some documentation (at least in the README.md) and add a test for this?

@andir Both done.

piegamesde:
What do you think of printing a trace warning for every override so that people don't accidentally run in to it?

thillux:
> What do you think of printing a trace warning for every override so that people don't accidentally run in to it?

Done.

Issue #113, opened 2025-02-05
error on add/import for implicit branch `main`


KiaraGrouwstra:
using npins 0.2.4, `npins import-flake`...

works for inputs with implicit branch `master`:

```nix
    flake-compat.url = "github:edolstra/flake-compat";
```

```
[INFO ] Importing flake-compat
```

yet for inputs with implicit branch `main` fails:

```nix
    flake-parts.url = "github:hercules-ci/flake-parts";
```

```
[INFO ] Importing flake-parts
[ERROR] Failed to import pin 'flake-parts'
[ERROR] Couldn't fetch the latest commit
[ERROR]         git ls-remote output is empty. Are you sure 'refs/heads/master' exists? Note: If you want to tag a revision, you need to also specify a branch ('--branch').
```

something similar happens when adding such entries separately (without specifying branch):

```
npins add github hercules-ci flake-parts
[INFO ] Adding 'flake-parts' …
Error: Failed to fully initialize the pin

Caused by:
    Repository has no matching release tags
```

(a similar symptom also seemed reported at https://github.com/andir/npins/issues/100#issuecomment-2338737236)

expected behavior: either default branch should just work

piegamesde:
CC @stuebinm

stuebinm:
oh dear.

afaict, there's two levels of unexpected things here: when I first wrote the flake importer, I assumes Nix would default to `master` rather than the repository's default branch, and that causes your `npins import-flake` to fail. Here, at least, the culprit [is obvious](https://github.com/andir/npins/blob/d9410c18ee4e6a8eb93bac6a1d5207f8350e767d/src/flake.rs#L103), as is the fix: npins should run `git ls-remote --symref <url> HEAD` to find the default branch, if none was given as an argument. Perhaps I'll find the time to fix that soon-ish, if nobody else does (but I'm a bit tight on free time currently, so no outright promises).

The second error is unrelated, subtler, and (to me) much more surprising: `npins` is [looking for a release tag](https://github.com/andir/npins/blob/d9410c18ee4e6a8eb93bac6a1d5207f8350e767d/src/git.rs#L517), rather than a branch, and finding none (which is, as such, correct, as that repository has none, whereas flake-compat does, and hence for that this works). But I'm not sure if this is also the intended behaviour?

piegamesde:
[...]

PR #114, opened 2025-02-08
import-flake: detect default branches correctly

stuebinm:
See https://github.com/andir/npins/issues/113

Previously, `npins import-flake` would simply assume that the branch was called `master` if not given, which is not correct; nix flakes will use the remote's default branch in such cases.

This mimics the behaviour in the flake importer — however, perhaps it might be better to move it into the Pin struct itself, allowing branch pins with no explicit branch name given, which then resolve to the remote's default branch instead?

### DIFF
--- a/src/flake.rs
+++ b/src/flake.rs
@@ -2,8 +2,8 @@
 
 use crate::*;
 use anyhow::{Context, Result};
+use git::fetch_default_branch;
 use serde::{Deserialize, Serialize};
-use std::convert::TryFrom;
 use url::Url;
 
 /// Pin entry from a nix flake's lock file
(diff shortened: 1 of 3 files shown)


KiaraGrouwstra:
thanks! this indeed addresses #113 for me :v:

KiaraGrouwstra (approved):

stuebinm:
I haven't thought about it in detail — perhaps we could have something like `npins add git <url> --default-branch`.

I'm not overlay attached to the idea and have little opinion on whether we want to have it or not, but I wanted to mention it here since it's a bit of a "low-hanging fruit" to add, should it be something people are interested in having (e.g. one might argue nix flakes do this, so we want this too, etc.)

src/flake.rs
```diff
@@ -60,60 +60,57 @@ impl FlakePin {
...
-                    "{}/{}",
-                    flake
-                        .locked
+        Ok(match self.locked.type_ {
+            Gitlab => {
+                // TODO: parsing the query string to retrieve servers other than
+                // gitlab.com is not supported for now, but could be added.
+                let branch = self.fetch_default_branch("https://gitlab.com").await?;
+                git::GitPin::gitlab(
+                    format!(
```
piegamesde, line 77: This is not new from this commit, but I absolutely don't like how the URL is mangled into the repo path here, whereas the actual URL field is then set to `None`. This seems to have been working fine so far, but it smells like a bug waiting to be discovered to me.

Moreover, doing it the "proper" way will give support for non-gitlab.com instances for free and should also make `fetch_default_branch` a bit simple. (Also yikes at the Flake lock format for very lazily retconning GitLab support onto a structure made for GitHub …)

stuebinm: i am not sure what exactly you mean? this `format!` mangles repository path & owner; on a flake input `gitlab:simple-nixos-mailserver/nixos-mailserver` this code produces a pin with `"repo_path": "simple-nixos-mailserver/nixos-mailserver"` without URL. iirc i left the URL parameter on `None` since it then using `https://gitlab.com` appeared to be the intended behaviour, but that could be changed; in either case, `GitPin::gitlab` will default it to `https://gitlab.com`, which'll be written into the `sources.json`.

The main reason i didn't add support for custom servers is that this'd require parsing the `?host=` query string of the flake input, which is its own can of worms (afaik, there seems to be no clear specification of which parts of a query string nix flakes will look at, and which if any will afterwards remain part of the actual url used for requests — tho this is less of a problem for gitlab urls, but for arbitrary git urls it can be relevant sometimes)

piegamesde: Oh gosh, this is awful :D
Thanks for the explanation

PR #115, opened 2025-02-25
Fixes and improvements

piegamesde:


test.nix
```diff
@@ -104,7 +104,7 @@ let
           npins
           python3
           netcat
-          nix
+          lix
```
andir, line 107: I'm fine with using `lix` for this but we should support running tests against `nix` as well. I don't think we have to default to brokenNix (better name than `CppNix` IMO) but there might be users and perhaps at some point we can add a smoke test as even `lix` might introduce bugs at some point.

src/git.rs
```diff
@@ -638,6 +639,10 @@ async fn fetch_remote(args: &[&str]) -> Result<Vec<RemoteInfo>> {
                 .unwrap_or_else(|| "None".into())
         );
     }
+    log::debug!("git ls-remote stdout:");
+    String::from_utf8_lossy(&process.stdout)
```
andir, line 643: While this probably works it wouldn't help a lot of `git` is stuck at some point during the execution. Since this is already `async` we could log lines as they appear? Probably something for the future but I'm usually deeply frustrated by applications that eat the error output (we only provide stdout) and that block until the subprocess is finished. Usually I just attach `strace` and check again, but nobody should be required to resort to that.

test.nix
piegamesde: Not sure how to best do this. `npins` has its `nix` dependency fixed in the derivation because of `wrapProgram $out/bin/npins --prefix PATH : "${runtimePath}"`, so even if the test fixture is flexible it still doesn't change what nix implementation npins uses for prefetching. Maybe we should just remove that and make it a "classic" run-time dependencey?

andir (approved):
:+1: looks good to me. A few minor things I'd change but those aren't really important right now. This is already a good improvement.

test.nix
andir: Maybe, I think in the past that might have helped keep the nix biniaries compatible with what we test against but ideally we'd use the fetching code of `tvix` anyway (or implement it on our own).

src/git.rs
piegamesde: I can look into it in a follow up. Ideally we'd factor out the process of shelling out into a helper function which does the appropriate wrapping, so that id doesn't have to be repeated at every call site (even though currently there are only two). In the case of git specifically, we could maybe also look into using libgit or any other Rust git bindings to call the code directly

Issue #47, opened 2023-07-31
Support for Gitea and other forges


piegamesde:
I somehow just remembered that there are other ways to host your software other than GitHub and GitLab. Of course we support all generic git repositories, nevertheless there are more forges which could benefit from the additional support.

Potential forges:
- Gitea
- ...?

Questions that need answering for each one:
- Is it self-hostable or fixed domain?
- Does it have `user/repo` structure for all repositories like GitHub or is it more free-form like GitLab?
- Does it host tarball artifacts for all commits?
  - Given a ref, what URL can its tarball be downloaded from?
  - Given a tag, what URL can its tarball be downloaded from?
- In anticipation of #44, given a release with uploaded artifacts, how to download them?

To generally keep the scope down, the current restrictions will also be required for new candidates:

- Finding appropriate versions is done through the generic Git API (`git ls-remote` etc.)
- Downloading files must be as simple as filling in an URL template and calling `curl`. Things that require adding a library for the specific API implementation are out of scope.

lf-:
I think the cleanest way to implement this in a maximally generic way is to implement Nix's immutable tarball protocol as used by flakehub and [as of yesterday, forgejo](https://codeberg.org/forgejo/forgejo/pulls/3615), which is a trivial amount of code to possibly just upstream into all the forges.

With this you can just give `https://some-forgejo/user/repo/archive/main.tar.gz`, which is a URL that does not need to be inspected at all.

Here is some hastily written Python I wrote to implement it today:

```python
import subprocess
import tempfile
from pathlib import Path
import re
import dataclasses
from typing import Literal
import urllib.parse
import json


@dataclasses.dataclass
class PinSerialized:
    kind: str
    rev: str | None
    nar_hash: str


@dataclasses.dataclass
class TarballPinSerialized(PinSerialized):
    kind: Literal['tarball']
    locked_url: str
    url: str


class PinSpec:

    def do_pin(self) -> dict[str, str]:
        raise ValueError('unimplemented')


@dataclasses.dataclass
class TarballPinSpec(PinSpec):
    url: str

    def do_pin(self) -> TarballPinSerialized:
        return lock_tarball(self.url)


@dataclasses.dataclass
class LinkHeader:
    url: str
    rev: str | None


LINK_HEADER_RE = re.compile(r'<(?P<url>.*)>; rel="immutable"')


def parse_link_header(header) -> LinkHeader | None:
    matched = LINK_HEADER_RE.match(header)
    if not matched:
        return None

[...]

Issue #44, opened 2023-07-03
Support specifying release artifact to prehash


fufexan:
I use npins in my [nix-gaming](https://github.com/fufexan/nix-gaming/) repo to automate updates. It's working great, but for some games I want to download a release artifact (compiled tarball/AppImage) instead of the source tarball. Since npins does not provide that functionality, I had to write update scripts for certain games (see [this](https://github.com/fufexan/nix-gaming/blob/b090e8b7e463d9c437536b25a0e9af3477a269e9/pkgs/osu-lazer-bin/update.sh)).

It would be great to have an option to specify the release artifact to get the hash for. Also, in case the artifact does not exist, the system could fallback to hashing the source tarball.

Another (maybe unrelated) feature request would be using SRI hashes, to be able to simply `inherit (pins.some-pin) hash` inside most fetchers.

piegamesde:
A release may have multiple uploaded artifacts, do I understand correctly that the idea is to provide the artifact name? And how should this interact across versions when updating?

> Also, in case the artifact does not exist, the system could fallback to hashing the source tarball.

That sounds weird to me. Usually, source and release have wildly different content, and simply swapping one for the other wouldn't help many people.

fufexan:
Yes, the idea is that we provide an artifact name that would be fetched from this URL:
`https://github.com/<owner>/<repo>/releases/download/<version>/<artifactName>`

> That sounds weird to me. Usually, source and release have wildly different content, and simply swapping one for the other wouldn't help many people.

In hindsight, it does sound weird. Nevermind the idea.

piegamesde:
What should happen on update if a release does not have a GitHub release, or the release does not have the desired artifact?

Maybe tangential, but it would be easy to create a pin type for fetching URLs, like a thin wrapper around `fetchTarball`. However, updating functionality would probably not exist.

fufexan:
> What should happen on update if a release does not have a GitHub release, or the release does not have the desired artifact?

I think erroring out is the best solution. If the artifact *should* exist but doesn't, then either upstream forgot about it, or there was a good reason to not include it anymore. Either way, it's the user's business from then on.

> However, updating functionality would probably not exist.

Why not? Currently, we can use the version provided by npins and use it for other fetchers. It's not exactly efficient when you don't need the source, but it does work.

piegamesde:
I fear that this may require adding support for artifact names which reference the current version to be useful, which does not sound like fun implementing :/

piegamesde:
[...]

PR #119, opened 2025-02-26
Add support for flakes "Lockable HTTP Tarball Protocol"

pr2502:
Hi, i've run into missing support for the "tarball" type of flake input when importing a flake. I've noticed it's been mentioned before in #47 so i gave it a try because it doesn't look very complicated.

This is just a quick experiment, if you want this this feature included in npins i'm happy to write tests and anything else that is missing^^

For now i've tested it only manually on my nixos config using
```sh
mkdir test && cd test
cargo run -- init --bare
wget https://git.p2502.net/max/nixconf/raw/branch/main/flake.lock
cargo run -- import-flake
cargo run -- show
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -62,9 +62,9 @@ pub struct ChannelAddOpts {
 }
 
 impl ChannelAddOpts {
-    pub fn add(&self) -> Result<(String, Pin)> {
+    pub fn add(&self) -> Result<(Option<String>, Pin)> {
         Ok((
-            self.name.clone(),
+            Some(self.name.clone()),
             channel::Pin {
                 name: self.name.clone(),
             }
@@ -119,9 +119,9 @@ pub struct GitHubAddOpts {
 }
 
 impl GitHubAddOpts {
-    pub fn add(&self) -> Result<(String, Pin)> {
+    pub fn add(&self) -> Result<(Option<String>, Pin)> {
         Ok((
-            self.repository.clone(),
+            Some(self.repository.clone()),
             match &self.more.branch {
                 Some(branch) => {
                     let pin = git::GitPin::github(
@@ -164,7 +164,7 @@ pub struct ForgejoAddOpts {
     pub more: GenericGitAddOpts,
 }
 impl ForgejoAddOpts {
-    pub fn add(&self) -> Result<(String, Pin)> {
+    pub fn add(&self) -> Result<(Option<String>, Pin)> {
         let server_url = Url::parse(&self.server).or_else(|err| match err {
             ParseError::RelativeUrlWithoutBase => {
                 Url::parse(&("https://".to_string() + self.server.as_str()))
@@ -173,7 +173,7 @@ impl ForgejoAddOpts {
         })?;
 
         Ok((
-            self.repository.clone(),
+            Some(self.repository.clone()),
             match &self.more.branch {
                 Some(branch) => {
                     let pin = git::GitPin::forgejo(
@@ -234,14 +234,14 @@ pub struct GitLabAddOpts {
 }
 
 impl GitLabAddOpts {
-    pub fn add(&self) -> Result<(String, Pin)> {
+    pub fn add(&self) -> Result<(Option<String>, Pin)> {
         Ok((
-            self.repo_path
+            Some(self.repo_path
                 .last()
                 .ok_or_else(|| anyhow::format_err!("GitLab repository path must at least have one element (usually two: owner, repo)"))?
-                .clone(),
+                .clone()),
             match &self.more.branch {
-                Some(branch) =>{
+                Some(branch) => {
                     let pin = git::GitPin::gitlab(
                         self.repo_path.join("/"),
                         branch.clone(),
@@ -249,11 +249,11 @@ impl GitLabAddOpts {
                         self.private_token.clone(),
                         self.more.submodules,
                     );
-                    let version = self.more.at.as_ref()
-                    .map(|at| git::GitRevision {
+                    let version = self.more.at.as_ref().map(|at| git::GitRevision {
                         revision: at.clone(),
                     });
-                    (pin, version).into()},
+                    (pin, version).into()
+                },
                 None => {
                     let pin = git::GitReleasePin::gitlab(
                         self.repo_path.join("/"),
@@ -264,10 +264,9 @@ impl GitLabAddOpts {
                         self.more.release_prefix.clone(),
                         self.more.submodules,
                     );
-                    let version = self.more.at.as_ref()
-                        .map(|at| GenericVersion {
-                            version: at.clone(),
-                        });
+                    let version = self.more.at.as_ref().map(|at| GenericVersion {
+                        version: at.clone(),
+                    });
                     (pin, version).into()
                 },
             },
@@ -285,7 +284,7 @@ pub struct GitAddOpts {
 }
 
 impl GitAddOpts {
-    pub fn add(&self) -> Result<(String, Pin)> {
+    pub fn add(&self) -> Result<(Option<String>, Pin)> {
         let url = Url::parse(&self.url)
             .map_err(|e| {
                 match e {
@@ -311,7 +310,7 @@ impl GitAddOpts {
         let name = name.strip_suffix(".git").unwrap_or(&name);
 
         Ok((
-            name.to_owned(),
+            Some(name.to_owned()),
             match &self.more.branch {
                 Some(branch) => {
                     let pin = git::GitPin::git(url, branch.clone(), self.more.submodules);
@@ -354,8 +353,8 @@ pub struct PyPiAddOpts {
 }
 
 impl PyPiAddOpts {
-    pub fn add(&self) -> Result<(String, Pin)> {
-        Ok((self.name.clone(), {
+    pub fn add(&self) -> Result<(Option<String>, Pin)> {
+        Ok((Some(self.name.clone()), {
             let pin = pypi::Pin {
                 name: self.name.clone(),
                 version_upper_bound: self.version_upper_bound.clone(),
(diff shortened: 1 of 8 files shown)


src/cli.rs
```diff
@@ -929,6 +929,7 @@ impl Opts {
                 .await
                 .context("Could not convert pin to npins format")?;
 
+            // QUESTION calling pin.update() here is surprising to me, shouldn't we try to reuse the hash from the flake in pin.try_to_pin() and only fetch here?
```
pr2502, line 932: This could've perhaps been an issue instead, but it's something i've run into it when writing the `try_to_pin()` impl for Tarball and it surprised me.

piegamesde: I don't know about import-flake, but when I wrote import-niv their hashing format was too much work to fully parse to extract the detailed version information, or maybe even incompatible in some areas (I forgot). Either way, focusing only on the pin inputs at the cost of an update was a reasonable compromise

piegamesde:
How does this works for tarballs which don't support that protocol? And would this solve the use case in #44?

pr2502:
as is currently it errors out about the missing header. i haven't read #44 before but i have been thinking about a similar use case and i think it could be solved by gracefully handling missing "original url" with `npins update` (log it's impossible to update and ignore) and adding cli options to add a locked url directly like `npins add tarball --locked <tarball url without the locking api support>` and something to manually specify an updated url like `npins set-url <input> --locked <tarball url without locking api support>`. i can add it to this PR or make a follow up^^

pr2502:
could be made the other way around, rename the current `url` field to `lockable_url` and `locked_url` to just `url` and make an tarball that cannot be automatically updated the default and move the lockable tarball api behind a flag like `npins add tarball --lockable <tarball url that supports auto updates via the link header>`

pr2502:
i think that's where i have ended up now? `npins add tarball <url>` does HTTP HEAD against the URL and if it returns the `Link: <flakeref>; rel="immutable"` header we store `<url>` as a `lockable_url` field (so we can later update the Pin) and the `flakeref` as `url`. if there is no such header we fall back to storing `<url>` as `url` and assuming it will always return the same thing

now what i'm missing is tests (i'm working on a python script to mock the header) and some command that would easily allow updating the tarballs without `lockable_url` (something like `npins set-url <name> <url>`.

i'd also like to propagate the information that a Pin cannot self update to the user interface of `npins update` so it shows as "requires manual update" instead of "unaltered"

pr2502:
i've finally figured out the integration tests in nix i think^^ the implementation i've tried before broke `npins update` for tarballs so i've fixed that too.

since `npins add` can be used twice with the same name (which i didn't know, thanks!) i don't need to implement anything for updating. and the `npins update` UI i'll leave alone until #78 to not introduce any more conflicts?

i'll mark this as ready for review now. should i clean up the commit history or just squash it into one commit?

src/cli.rs
```diff
@@ -368,6 +368,33 @@ impl PyPiAddOpts {
...
 
+#[derive(Debug, Parser)]
+pub struct TarballAddOpts {
+    /// Lockable URL
+    pub url: Url,
+}
+
+impl TarballAddOpts {
+    pub fn add(&self) -> Result<(String, Pin)> {
+        // TODO this assumes common forgejo/github/... URL scheme of:
```
pr2502, line 379: this is a bit i'm really not sure about, what do you think?

piegamesde: Yeah I already stumbled upon this in the earlier draft, forgot to comment on it. I'm a bit torn, because it will kinda work as a good default most of the time, but when it won't it will be confusing and generally a bad experience. Therefore I'm inclined to remove that part and always return `None` for the name instead

src/tarball.rs
```diff
@@ -0,0 +1,153 @@
...
+                    "rev" => {
+                        revision = Some(val.to_string());
+                    },
+                    _ => {}, // Ignore other query pairs
+                }
+            }
+            (Some(tarball_url), locked_url)
+        } else {
+            if let Some(lockable_url) = old.and_then(|old| old.lockable_url.as_ref()) {
+                log::warn!("lockable tarball url `{lockable_url}` did not respond with the appropriate `Link` header");
```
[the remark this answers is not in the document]
piegamesde: Not sure, maybe? Can I reasonably expect a URL locked with that protocol to always be provided under that protocol? But also, if we error out we should provide a way to tell the user how to fix their pin

piegamesde:
I merged the freeze pins PR so you aren't blocked on it. Though I am unsure if going down that route is still the best idea

src/cli.rs
pr2502: ok, i can definitely do that. is it ok if i change all the other function impls to return `Option<String>` instead of `String` too to make it consistent, or should i rather work around this one being out of line [here](https://github.com/andir/npins/pull/119/files#diff-b2812f19576dd53d0c35b107a322f58a00fce6977f4b5976e1961853982af3ccR441-R448)?

piegamesde: Technically, the downstream caller checks for an empty string, so simply passing `"".into()` would technically work. Doing a small refactoring would be cleaner, but I can do it as a follow-up instead if you prefer. It's not urgent

src/tarball.rs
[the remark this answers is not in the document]
pr2502: i've left the warning with more explanation and made it exit successfully without modifying the Pin. i wouldn't expect this to be a common case, this protocol is supported by flakehub and relatively recently Jade added it to forgejo. so it could happen if someone downgraded their forgejo instance or maybe switched to a different forge software on the same domain

pr2502:
i know this is off topic on this PR but looking closer at the Pin freezing code after i've rebased on it i've noticed it's going to add `"frozen": false` to every pin on update and cause a lot off diff noise, could we add a `#[serde(skip_serializing_if = "is_not_frozen")]` to the field?

src/cli.rs
pr2502: changed it to always ask and did the small refactor for the rest of them^^

piegamesde:
Neat, would you mind extracting that commit into its own PR?

pr2502:
sure^^ it's here https://github.com/andir/npins/pull/122

src/tarball.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+        if let Some(revision) = &self.revision {
+            props.push(("revision".into(), revision.clone()))
+        }
+        props
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+pub struct TarballHash {
```
piegamesde, line 59: This could be moved into the top-level module as `GenericHash`

src/tarball.rs
```diff
@@ -0,0 +1,166 @@
...
+pub struct TarballPin {
+    /// URL provided as user input.
+    ///
+    /// Will get moved into `LockedTarball` once the Pin has been updated at
+    /// least once. This is a little bit of a hack but we need to be able to
+    /// update the `url` from within `TarballPin::update` so we have to store it
+    /// within `LockedTarball` which doesn't exist yet the first time `update`
+    /// is called.
+    #[serde(skip)]
+    pub url: Option<Url>,
```
piegamesde, line 23: I very much do not like this, what are the alternative designs? Would it be possible to simply always store that url regardless of locking?

You have `TarballPin { url }` which always contains the input URL. Then you have `LockedTarball { Option<locked_url> }` if the `url` is lockable. The fetcher uses `locked_url` if present and `url` otherwise.

src/tarball.rs
```diff
@@ -0,0 +1,166 @@
...
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
+pub struct LockedTarball {
+    /// If the given URL supports the Lockable Tarball Protocol we use the
+    /// locked one for `url` and we store the original used for updating here
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub lockable_url: Option<Url>,
+    /// Either the URL provided by user or an immutable URL returned in Link header
+    pub url: Url,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub revision: Option<String>,
```
piegamesde, line 41: Given that we're not implementing the full Flake URL specification and are using a plain tarball fetcher, maybe let's skip these for now altogether.

src/cli.rs
```diff
@@ -759,6 +777,7 @@ impl Opts {
             .filter(|(_, (name, pin))| {
                 (opts.names.is_empty() || valid_names.contains(name))
                     && (opts.update_frozen || !pin.is_frozen())
+                    && pin.can_auto_update()
```
piegamesde, line 780: It turns out this logic is wrong, because it does not account for the update strategy. While such a pin cannot *update* in the sense of `Updatable::update`, it can still have a hash change in the sense of `Updatable::fetch`, which is handled in the same code path.

Moreover, given that we cannot guarantee information on a tarball pin which we do not have (e.g. before first fetch) or if the server implementation changes its mind, my suggestion would be to remove the `can_auto_update` logic again and simply let the pin do its thing (meaning that regular tarball pins will simply happen to be up to date all of the time).

piegamesde (changes requested):
Sorry for the back and fourth on the design, it's an iterative process

src/tarball.rs
pr2502: yes, definitely agreed. i don't know why i got stuck in the mindset that we always have to put the url to download into `url`. adding a condition to the fetcher is much much simpler

src/cli.rs
pr2502: true, i didn't realize that. i've just reverted the whole commit^^

src/tarball.rs
pr2502: oki, i've removed the whole bit

src/default.nix
```diff
@@ -129,9 +129,14 @@ let
     };
 
   mkTarballSource =
-    { url, hash, ... }:
+    {
+      url,
+      locked_url ? null,
```
piegamesde, line 134: You can do `locked_url ? url` and then use `url = locked_url` below

src/versions.rs
```diff
@@ -5,7 +5,7 @@ use anyhow::{Context, Result};
 use serde_json::{json, Map, Value};
 
 /// The current format version
-pub const LATEST: u64 = 4;
+pub const LATEST: u64 = 5;
```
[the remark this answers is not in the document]
pr2502: thanks! i've completely missed that

pr2502:
i've just squashed everything into one commit, i don't think it's complicated enough to warrant splitting up. hopefully now without typos :D

piegamesde (approved):

PR #120, opened 2025-02-28
cli: Add lockfile mode

### LINKED ISSUE #105: Consider not requiring checking in generated code
Currently, `npins/default.nix` is generated by npins for projects using it. I wonder if it would be possible to fetch npins itself as a library from `npins/sources.json` in `default.nix` with a small amount of code and call into the fetched library to do the work instead. `flake-compat` works this way, for example.

piegamesde:
This supersedes #30 and #48 and closes #105. The design is a bit clunky, but unlike #48 it allows to fully customize the sources.json file name. In the end, everything is a compromise

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,10 +116,11 @@ Commands:
   help          Print this message or the help of the given subcommand(s)
 
 Options:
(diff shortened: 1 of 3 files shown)


piegamesde:
How does this works for tar balls which don't support that protocol? And would this solve the use case in #44?

PR #121, opened 2025-03-03
Libnpins

piegamesde:


piegamesde:
This totally isn't ready for actual lib usage yet, however I'd still like to get this merged to avoid conflict rot because of the refactorings necessary to do this

andir:
Lets merge it after the current release and then we can iterate on the API interface.

PR #127, opened 2025-03-30
cli: Fix reported file name

samueldr:
There may be more to do, now that *lockfile mode* is a thing. The file name may be different, in other code paths, too.


piegamesde:
The instance you found is way older than lockfile mode from a very early refactoring, but now that you've said this lockfile mode has the same problem and I even found a major bug in `upgrade` …

samueldr:
Yes, I was tired when I wrote the description yesterday.

This changes this one old instance... but *now that lockfile mode is a thing*, there are a lot more to think about regarding reporting the file name.

PR #132, opened 2025-04-09
fix the source name when calling `nix-prefetch-url`

fricklerhandwerk:
since by default `nix-prefetch-url` derives the symbolic store path name from the URL, but `builtins.fetchTarball` does not, setting `--name source` avoids downloading a source twice:
1. when running `npins add`
2. when evaluating `default.nix`

Closes https://github.com/andir/npins/issues/57 based on @roberth's suggestion in https://github.com/NixOS/nix/issues/10881#issuecomment-2573507168

This change is transparent for existing users of `npins`: The symbolic store path name is only different from previous versions *on first download* when running `npins add` or `npins update`, which would produce a new store path anyway.

### DIFF
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -10,11 +10,13 @@ pub struct PrefetchInfo {
 pub async fn nix_prefetch_tarball(url: impl AsRef<str>) -> Result<String> {
     let url = url.as_ref();
     log::debug!(
-        "Executing `nix-prefetch-url --unpack --type sha256 {}`",
+        "Executing `nix-prefetch-url --unpack --name source --type sha256 {}`",
         url
     );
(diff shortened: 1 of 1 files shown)


piegamesde:
How does this compare to #74?

fricklerhandwerk:
It's much simpler because it doesn't do the dance of mimicking behavior that's IMO completely unnecessary. In fact, here's a Nixpkgs patch to remove the need for that other hack: https://github.com/fricklerhandwerk/nixpkgs/commit/1412966acb0b6c6a10e0290b62e958c76b0a4224

KiaraGrouwstra:
would it be possible to do a new version tag for this? :pray:

piegamesde:
That's the plan, but I have a couple more bugs to squash first

PR #133, opened 2025-04-16
Towards a new release

piegamesde:


src/cli.rs
```diff
@@ -813,15 +827,19 @@ impl Opts {
...
             .await?
             .into_iter()
-            .for_each(|(name, diff)| print_diff(name, diff));
+            .for_each(|(name, diff)| {
+                has_diff |= !diff.is_empty();
+                print_diff(name, diff);
+            });
 
-        if !opts.dry_run {
+        if has_diff && !opts.dry_run {
```
Coca162, line 842: Since the empty differences are already filtered out you could do it like this instead:
```rust
        let differences = stream::iter(update_iter)
            .buffer_unordered(opts.max_concurrent_downloads)
            .try_filter(|(_, diff)| future::ready(diff.is_empty().not()))
            .try_collect::<Vec<_>>()
            .await?;

        for (name, diff) in &differences {
            print_diff(name, diff);
        }

        if !differences.is_empty() && !opts.dry_run {
```

src/cli.rs
```diff
@@ -793,10 +807,10 @@ impl Opts {
                 anyhow::Result::<_, anyhow::Error>::Ok((name, diff))
             });
 
-        stream::iter(update_iter)
+        let differences = stream::iter(update_iter)
             .buffer_unordered(opts.max_concurrent_downloads)
-            .try_for_each(|(name, diff)| {
-                fn write_diff(writer: &mut impl Write, name: &str, diff: Vec<diff::DiffEntry>) {
+            .try_filter(|(name, diff)| {
```
Coca162, line 812: Since we don't use them afterwards now you could just use [try_any](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_any) and rename the var to `has_diff` like it was before.

piegamesde: `try_any` is neat. I had looked for `try_count` which doesn't exist but since I always want to check non-empty this makes more sense

piegamesde: Oh wait no. `try_any` short-circuits which we absolutely do not want here

Coca162: argh! Maybe just return to just having a mutable bool? I think it's a bit weird to collect a vec just to not use the values inside them

Coca162: Though *could* also use `try_all` since you can invert the logic here to instead have like a `no_diffs`.

piegamesde: I've thought of that, but `try_all` also short-circuits. After a brief detour over `try_fold` which turned out to be like a `mut has_diff` but strictly worse, I'm indeed now back to the mutable bool soltuion. Also has the shortest diff anyways

PR #134, opened 2025-04-19
document using pins thru `NIX_PATH`

KiaraGrouwstra:
inspired by [jade's blog post](https://jade.fyi/blog/pinning-nixos-with-npins/) on using npins with NixOS, i played around with exposing pinned packages beside `nixpkgs` as well.

in practice, one may want to also set its `nixos-config` or further expose `flake` to the path (following jade's post), but i tried to kind of scope down my explanation to the part i considered most relevant for npins here.


KiaraGrouwstra:
(potentially even more fun is exposing npins [processed by `readTree`](https://cl.snix.dev/c/snix/+/30330) :no_mouth:)

samueldr:
The main worry with using `NIX_PATH` is that this is causing ambient impurities to exist *sometimes* in relation to the environment in which the eval is running.

While the angled bracket syntax is neat, and convenient, having it be implemented using `NIX_PATH` has the same broad usability concerns as usual.

BTW, not saying this is undesirable. I'm more concerned about how this should be more explicit in the documentation.

For example, this means that this is not proper declarative outputs for an expression, once angled brackets are in use. As an example, when running `npins update` *after* having used the `jq` snippet, the current ambient environment will still refer to the previously evaluated pins.

(:thinking: This one specific usability concern could be helped by making use of a `nix-profile`-esque scheme, but that's also even worse and more cursed...)

As long as the scheme to resolve a pin lies outside of the evaluation of an expression, you're at risk of referring to stale data, or data that hasn't been loaded, or worse, data from some other ambient environment.

* * *

Sorry, this is a lot to unpack. Does that make sense? And note that it's pretty much in no way related to your suggested changes, it's the broader problem related to `NIX_PATH`.

samueldr:
One way this *could* be helped is by having `npins` aware of Nix operations and handling it transitively.

I.e. something like `npins nix-path-exec nix-build '<nixpkgs>' -A hello`, which would setup the `NIX_PATH` entry (clobbering it entirely, or extending it?), and child processes would use it...

... except that the same broader usability issues still exist: if running something that is not directly a user of `NIX_PATH`, or something that runs additional processes, like you would a `nix-shell`, this will cause environmental pollution.

So it might be better in this case to have helpers that are more aware for this, for using `nix-shell`.

I wonder what semantics such a thing should have:

```
npins shell [???]
```

It probably shouldn't limit itself to `<nixpkgs>`-isms like `nix-shell` does, but it also kind of needs to be compatible with it for better integration... bleh the weight of legacy surrounding this area is intense. And not helped [by how Flakes have de facto blocked most other efforts](https://samuel.dionne-riel.com/blog/2023/09/06/flakes-is-an-experiment-that-did-too-much-at-once.html).

KiaraGrouwstra:
thanks for your feedback @samueldr. i pushed an update to hopefully be a bit more explicit about these concerns. does that look better to you?

wrapping this into an npins command i feel would make it a bit less transparent how the variable is set - potentially relevant so the user may still know how to add further variables there, `nixos-config` or `flake` included.

as for flakes, i feel like we're making progress posing a viable alternative. :)

piegamesde:
The usual issues with $NIX_PATH aside, in my eyes the documentation you propose is too specific on a specific use case for npins. Npins is meant to be a source of inputs for all Nix projects and dev shells, where system configuration are just one special case of that. For local dev shells, setting `$NIX_PATH` would be acceptable because they get regenerated on every enter, but at the same time the naive approach leads to a bootstrapping problem that likely requires an external tool like direnv to resolve (I can't fully say, because I've only ever used `nix-shell` and can't tell how other Nix-based dev shells work here). As for the system wide configuration, due to the way environment variables propagate, the `NIX_PATH` would become stale with every update until the next relogin. Jade's blog posts handles that, but it's hacky to a degree that I wouldn't want to make it "official" in our documentation.

-------

> (:thinking: This one specific usability concern could be helped by making use of a nix-profile-esque scheme, but that's also even worse and more cursed...)

I need that, and it probably wouldn't be much cursed than abusing a pseudo flake registry for the same purpose like Jade did.

KiaraGrouwstra:
> at the same time the naive approach leads to a bootstrapping problem that likely requires an external tool like direnv to resolve

hm, i think the bash/jq example both does not suffer from the bootstrapping issue (unlike setting from nixos without that), on top of being general to nix rather than nixos-specific?

because then i can just leave out the whole nixos bit - see new push.

piegamesde:
This PR and @samueldr's comment sent me on a two week deep rabbit hole, which resulted in this: https://piegames.de/dumps/pinning-nixos-with-npins-revisited/.

Moreover, I'm currently implementing `npins get-path` (no PR yet) to get the path of pins without having to resort to `nix-instantiate` and `jq`.

piegamesde:
Coming back to this, what about instead of having the environment variable one did the same thing in Nix instead? `__nixPath = builtins.attrValues (builtins.mapAttrs (name: val: { path = val.outPath; prefix = name; }) (import ./npins))` Making this propagate through imports requires another cursed incantation of similar kind, but is definitely feasible

KiaraGrouwstra:
@piegamesde: what did you wanna propagate? i just put `sources` in `specialArgs`.

on using nix: i had originally been using both the nix approach (to resolve lookup paths after build) as well as the env var (to resolve lookup paths during the build), but to be fair, i lost interest in using lookup paths during build at this point.

originally i considered them nice syntax sugar, tho since then i started considering them a footgun, as their syntax makes it easy to resolve sub-paths, whereas in practice it would then just separately copy those to the store, rather than reusing their root.

but yeah, maybe just documenting the nix part could be useful to some people.

piegamesde:
I think I'd like to close this, such hacks are nice but probably have a better place on personal blog sites

PR #138, opened 2025-05-02
Rework update UI to accomodate larger pin lists

### LINKED ISSUE #136: Update command looks weird with many pins
This was unlikely to occur back when the UI was introduced in #112, but now that it is possible to use npins as a generic tool with lockfile mode, lock files can easily have hundreds of pins. As soon as the number of pins overflows the terminal height, glitches start to occur.

Reproducer: `npins --lock-file test.json update --dry-run`

[test.json](https://github.com/user-attachments/files/19864220/test.json)

CC @Coca162

-----

I'm open as to how to best fix this; A more sophisticated rendering which takes the terminal height into account might be feasible but not necessarily required. Maybe the simplest fix would be to not print queued pins and then have a sliding window of only the currently fetching pins, which would have only the size of the number of concurrent downloads which is a lot smaller.

Coca162:
Closes #136

Still need to work on having a progress bar/counter for how many pins are left to update as well as non-terminal support.  

https://github.com/user-attachments/assets/97751699-0cfc-4139-a45f-92b7c37d871d

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,21 +1,23 @@
 //! The main CLI application
-use std::collections::BTreeMap;
 use std::path::PathBuf;
+use std::{cell::RefCell, collections::BTreeMap};
 
 use npins::*;
 
 use std::{
     collections::BTreeSet,
-    io::{stderr, stdout, IsTerminal, Write},
+    io::{stderr, Write},
     ops::Not,
 };
 
 use anyhow::{Context, Result};
 use clap::{Parser, Subcommand};
 use crossterm::{
-    cursor, execute,
-    style::{ContentStyle, Print, StyledContent, Stylize},
-    terminal,
+    cursor::MoveToPreviousLine,
+    queue,
+    style::{Print, Stylize},
+    terminal::{Clear, ClearType},
+    QueueableCommand,
 };
 use futures::{
     future,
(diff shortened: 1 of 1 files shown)


Coca162:
With a counter now:

https://github.com/user-attachments/assets/31d73922-dc93-4a3f-bf8d-19d1f356fb76

And how it looks in non terminals:

https://github.com/user-attachments/assets/7b5c9bc5-29d8-4106-a150-5e5ce669ea80

Might in a later PR make the counter a progress bar, but this should be usable enough as is.

src/cli.rs
```diff
@@ -731,99 +719,132 @@ impl Opts {
...
+            })
+            .await?;
+        if length != 0 && stderr().is_terminal() {
+            eprintln!();
+        }
 
         if !opts.dry_run {
             self.write_pins(&pins)?;
-            log::info!("Update successful.");
+            log::info!("Updated lockfile.");
```
piegamesde, line 845:
```suggestion
            log::info!("Updated successful.");
```

Any strong reason for that change? Otherwise I'd prefer it the current way

src/cli.rs
```diff
@@ -731,99 +719,132 @@ impl Opts {
...
+
+                in_progress.remove(name.as_str());
+                write_diff(&mut stderr, name, diff);
+
+                for n in in_progress.iter() {
+                    stderr.queue(Print(n.dark_yellow())).unwrap();
+                    stderr.write_all(b"\n").unwrap();
+                }
+                finished.set(finished.get() + 1);
+                write!(stderr, "{}/{length} Pins", finished.get()).unwrap();
```
piegamesde, line 833:
```suggestion
                write!(stderr, "Updated {}/{length} pins", finished.get()).unwrap();
```

src/cli.rs
Coca162: I think it makes it clearer what the condition is behind the log, but that's not really strong reasoning since users should only really care about this if they are specifically thinking about dry runs.

src/cli.rs
```diff
@@ -731,99 +719,132 @@ impl Opts {
...
+        let finished = Cell::new(0);
 
         let update_iter = pins
             .pins
             .iter_mut()
-            .enumerate()
-            .filter(|(_, (name, pin))| {
-                (opts.names.is_empty() || valid_names.contains(name))
-                    && (opts.update_frozen || !pin.is_frozen())
+            .filter(|(name, pin)| {
```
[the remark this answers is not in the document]
Coca162: To create a `BTeeMap<&str, &mut Pin>` you would have to call `pins.pins.get_mut(name)` multiple times which rust doesn't like so it would basically require the same logic with `pins.pins.iter_mut()` but now handling the update all case has to also allocate into `selected_pins`.

piegamesde (approved):

PR #139, opened 2025-05-06
feat: Use narHash for Git sources

Tom-Hubrecht:
It is a follow-up of #87 using snix nix-compat crate for manipulating hashes.

The upgrade also takes care of migrating `GitRelease` pins, and is now done incrementally

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -82,7 +82,7 @@ let
     if url != null && !submodules then
       builtins.fetchTarball {
         inherit url;
-        sha256 = hash; # FIXME: check nix version & use SRI hashes
+        sha256 = hash;
       }
     else
       let
@@ -109,9 +109,9 @@ let
       in
       builtins.fetchGit {
         rev = revision;
-        inherit name;
-        # hash = hash;
-        inherit url submodules;
+        narHash = hash;
+
+        inherit name submodules url;
       };
 
   mkPyPiSource =
@@ -140,7 +140,7 @@ let
       sha256 = hash;
     };
 in
-if version == 5 then
+if version == 6 then
   builtins.mapAttrs mkSource data.pins
 else
   throw "Unsupported format version ${toString version} in sources.json. Try running `npins upgrade`"
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -1,11 +1,22 @@
 use crate::check_url;
 use anyhow::{Context, Result};
+use data_encoding::BASE64;
 use log::debug;
 
 #[allow(unused)]
 pub struct PrefetchInfo {
     store_path: String,
     hash: String,
+}
+
+pub fn hash_to_sri(s: &str, algo: &str) -> Result<String> {
+    let hash = nix_compat::nixhash::from_str(s, Some(algo))?;
+
+    Ok(format!(
+        "{}-{}",
+        hash.algo(),
+        BASE64.encode(hash.digest_as_bytes())
+    ))
 }
 
 pub async fn nix_prefetch_tarball(url: impl AsRef<str>) -> Result<String> {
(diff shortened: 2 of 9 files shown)


piegamesde:
I'm slightly confused about the nature of this PR, probably because I don't understand what it's doing too deeply.

- Does this mean that npins now uses SRI hashes everywhere?
- Why does the code only change the git fetching code path, yet also touches on pins that are git-release and not using any git fetchers?
- What do NAR hashes have to do with this?

Tom-Hubrecht:
> I'm slightly confused about the nature of this PR, probably because I don't understand what it's doing too deeply.
> 
>     * Does this mean that npins now uses SRI hashes everywhere?

No, only for locking `Git` and `GitRelease` pins, although, it would be possible to make all pins use SRI hashes.
Actually, it should be made so that every pin use SRI hashes for consistency.

>     * Why does the code only change the git fetching code path, yet also touches on pins that are git-release and not using any git fetchers?

So, git-release pins use direct git fetching if submodules are enabled or if no url is present for the tarball.

>     * What do NAR hashes have to do with this?

It is the way CppNix (at least between 2.19 and 2.24) and Lix (when https://gerrit.lix.systems/c/lix/+/3009 is merged) internally consider inputs from builtins fetchers locked and will thus try to look them in the store or substitute them first instead of always fetching the remote repository

piegamesde:
Okay, thank you for the explanation. Would you mind implementing the SRI hashes everywhere?
Will the code still work on Lix versions without that patch merged? Also, please drop the chore commits and rebase, I recently updated all dependencies for the 0.3.1 release.

Tom-Hubrecht:
> Will the code still work on Lix versions without that patch merged?

Yes, the `narHash` attribute is always accepted in `builtins.fetchGit` but is unused ( except for specific conditions c.f. https://github.com/nikstur/lon/pull/3 )

src/default.nix
```diff
@@ -82,7 +82,7 @@ let
     if url != null && !submodules then
       builtins.fetchTarball {
         inherit url;
-        sha256 = hash; # FIXME: check nix version & use SRI hashes
+        sha256 = hash;
```
piegamesde, line 85: Can't we then also switch to `inherit hash` now? (same for all other `fetchTarball` invocations)

src/nix.rs
```diff
@@ -1,13 +1,25 @@
 use crate::check_url;
 use anyhow::{Context, Result};
+use data_encoding::BASE64;
 use log::debug;
+use nix_compat::nixhash::from_str;
```
piegamesde, line 5: Slight nit: Importing a function with such a generic name as `from_str` and that for just one usage; I'd prefer simply using the fully qualified name at call site instead.

src/versions.rs
```diff
@@ -224,6 +236,46 @@ fn upgrade_v0_pin(name: &str, raw_pin: &mut Map<String, Value>) -> Result<()> {
     Ok(())
 }
 
+/* v5→v6. This upgrade changes the hashes of git and git-release pins to use SRI hashes instead of
+ * raw sha256 hashes.
+ */
+fn upgrade_v5_pin(name: &str, raw_pin: &mut Map<String, Value>) -> Result<()> {
+    let pin: Pin = serde_json::from_value(serde_json::Value::Object(raw_pin.clone()))?;
```
piegamesde, line 243: This is an acceptable solution for now, however it is not entirely future proof as the `Pin` type used here will implicitly bitrot with following migrations. This can be worked around with the next migration, however, if you could find a stringly-typed solution that simply replaces "hash" fields that would be favorable.

(The general migration solution is still lacking, however the only way I could see building a fully typed solution would be by duplicating the entire `Pin` type with all its variants for every version migration and ugh)

src/default.nix
Tom-Hubrecht: Well, the interface of te builtins fetchers is all over the place, `fetchTarball` and `fetchurl` accept only `url`, `name` and `sha256` (which can be plain, or in SRI form), while `fetchGit` uses the `fetchTree` interface which is different and uses `narHash` for the hash

PR #140, opened 2025-05-09
Refactoring and fixes

piegamesde:
This PR is best reviewed commit by commit.

This is mostly refactoring stuff, one very minor bug fix and some slight improvements to the error handling.

CC @Tom-Hubrecht I saw in your PR that the old version upgrading code was insufficient for our new needs, however I wanted something more future proof than your approach (also yours would probably lead to a couple of confusing log messages). TPAL

### DIFF
--- a/src/versions.rs
+++ b/src/versions.rs
@@ -54,38 +54,57 @@ pub fn upgrade(mut pins_raw: Map<String, Value>) -> Result<Value> {
             )
         })?;
 
-    /* This is where the upgrading happens (at the moment we don't have any versions to upgrade from) */
-    match version {
-        0 => {
-            let pins = pins_raw
-                .get_mut("pins")
-                .and_then(Value::as_object_mut)
-                .ok_or_else(|| anyhow::format_err!("sources.json must contain a `pins` object"))?;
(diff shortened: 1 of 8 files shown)


src/versions.rs
```diff
@@ -54,38 +54,63 @@ pub fn upgrade(mut pins_raw: Map<String, Value>) -> Result<Value> {
...
+            version
+        );
+    } else if version == LATEST {
+        log::info!("sources.json is already up to date");
+    } else if version_upgraders.range(version..).next().is_none() {
+        /* We are not at the latest version, but there is no upgrader registered, which means that
+         * the upgrade is already being handled by serde default fields. The diff will always be
+         * the bump of the version field and nothing more.
+         */
+        log::info!("There is nothing to do");
```
Tom-Hubrecht, line 107: I feel like this test is not necessary, and could be replaced with
```rust
log::info("Upgrade complete");
```
after the `for` loop

piegamesde: I migrated this code path out of inertia, but you're right that it isn't really needed. I'll remove it

PR #145, opened 2025-05-23
add pinning container images

### LINKED ISSUE #126: Support for container images
As the title suggest, support for container images via pullImage (or maybe even buildImage) would be nice. Not exactly sure if this project is interested in supporting container images but if it is, please let me know!

cramt:
fixes #126

adds support for pinning OCI containers with the help of the `nix-prefetch-docker` command

### DIFF
--- a/src/nix.rs
+++ b/src/nix.rs
@@ -125,3 +125,69 @@ pub async fn nix_prefetch_git(
     };
     check_git_url(result.await, url).await
 }
+
+#[allow(unused)]
+#[derive(Debug, serde::Deserialize)]
+pub struct NixPrefetchDockerResponse {
+    pub hash: String,
+    #[serde(rename = "imageName")]
+    pub image_name: String,
+    #[serde(rename = "imageDigest")]
+    pub image_digest: String,
+    #[serde(rename = "finalImageName")]
+    pub final_image_name: String,
+    #[serde(rename = "finalImageTag")]
+    pub final_image_tag: String,
+}
+
+pub async fn nix_prefetch_docker(
+    image_name: impl AsRef<str>,
+    image_tag: impl AsRef<str>,
+    image_digest: Option<&str>,
+) -> Result<NixPrefetchDockerResponse> {
+    let image_name = image_name.as_ref();
+    let image_tag = image_tag.as_ref();
+
+    log::debug!(
+        "Executing: `nix-prefetch-docker {} {}{}`",
+        image_name,
+        image_tag,
+        match image_digest {
+            Some(x) => format!(" --image-digest {}", x),
+            None => "".into(),
+        }
+    );
+    println!("{} {}", image_name, image_tag);
+    let mut output = tokio::process::Command::new("nix-prefetch-docker");
+    let output = output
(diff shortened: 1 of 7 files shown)


piegamesde:
Looking good on a first glance, though still needs some tests. Also the `default.nix` needs updating, and this requires a format version bump.

src/default.nix
```diff
@@ -139,8 +150,21 @@ let
...
     };
+
+  mkContainerSource =
+    {
+      image_name,
+      image_tag,
+      image_digest,
+      ...
+    }:
+    pkgs.dockerTools.pullImage {
```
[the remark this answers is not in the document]
cramt: sadly not, i did go looking for a while hoping to find something when i was working on this, i kinda hoped this was okay, but its clearly not, we could pin a channel hash along with container, or maybe we could have a single "main" channel hash used for internal npins things, called `npins_internal` or something?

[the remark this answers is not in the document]
cramt: having a "main" channel hash for internal npins things might be the best future proofing solution since other more ecosystem specific fetchers also depend on nixpkgs like the ruby one fx?

cramt:
Setting up some full integration tests have been surprisingly simple, currently im only stumped on the fact that nix-preftech-docker *requires* https which is annoying

cramt:
we got there chief :saluting_face:, sadly i had to add some artifacts needed for the integration tests, specifically an image and also some ssl stuff because nix-prefetch-docker is annoying

piegamesde:
time and energy for review :(

src/nix.rs
```diff
@@ -127,6 +127,71 @@ pub async fn nix_prefetch_git(
...
+    log::debug!(
+        "Executing: `nix-prefetch-docker {} {}{}`",
+        image_name,
+        image_tag,
+        match image_digest {
+            Some(x) => format!(" --image-digest {}", x),
+            None => "".into(),
+        }
+    );
+    println!("{} {}", image_name, image_tag);
```
piegamesde, line 161: please remove this or use a logger instead here

test_artifacts/cert.pem
```diff
@@ -0,0 +1,13 @@
+-----BEGIN CERTIFICATE-----
```
piegamesde, line 1: I think I'd rather have the contents of the `test_artifacts` folder be simply in `tests`, or maybe `tests/assets`

src/default.nix
```diff
@@ -182,6 +196,23 @@ let
...
+  mkContainerSource =
+    pkgs:
+    {
+      image_name,
+      image_tag,
+      image_digest,
+      ...
+    }:
+    if pkgs == null then
+      builtins.throw "container sources requires nixpkgs"
```
piegamesde, line 209:
```suggestion
      builtins.throw "container sources require passing in a Nixpkgs value: https://github.com/andir/npins/blob/master/README.md#using-the-nixpkgs-fetchers"
```

piegamesde (changes requested):
The npins specific stuff looks good to me, though I don't use containers so can't really review that in-depth.

PR #150, opened 2025-06-16
git: Fix GitLab support

### LINKED ISSUE #146: gitlab archive URL doesn't retrieve correct ref
Found by @ccicnce113424 https://github.com/fufexan/nix-gaming/pull/259/files#r2107449550

This archive URL downloads HEAD, not ref=v2.1, so npins is pinning HEAD while calling it v2.1. It'll presumably break when next pushed due to hash mismatch.

Workaround is to treat it as a generic git repo rather than gitlab.

https://github.com/andir/npins/blob/eed27459effed464376369c9ab7e2972889fb821/src/git.rs#L266

piegamesde:
Turns out it has always been fundamentally broken. Not anymore (hopefully)

Fixes #146

CC @LunNova

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -263,7 +263,7 @@ impl Repository {
                         ]
                         .iter(),
                     );
-                url.set_query(Some(&format!("ref={}", tag)));
+                url.set_query(Some(&format!("sha={}", tag)));
                 if let Some(token) = private_token {
                     url.set_query(Some(&format!("private_token={}", token)));
                 }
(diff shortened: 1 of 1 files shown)


piegamesde:
I still have quite a few other things planned for 0.4, and also 0.4 wouldn't be backported to NixOS stable, so I'll try to cut a patch release in the next couple of weeks

christianlask:
Awesome :) Thank you very much.

PR #157, opened 2025-07-20
Rewrite update command with progress bars

Coca162:
I wanted to do this with a hand written implementation but it became rather messy and so I brought in `indicatif` instead to replace the code.

https://github.com/user-attachments/assets/fe7fa3f7-288e-41a6-ac65-a5bcd4b9506f

The no terminal version also looks exactly the same as previously, though now `--verbose` disables the animation since logs don't work with it (this was true previously but just not dealt with).


piegamesde:
Uhm. Can you please look at <https://github.com/piegamesde/npins/commit/2fc0caddc3ebb680b33d8c9475eacaf459dc5b53>? What do you think of that design?

Coca162:
@piegamesde I had considered properly abstracting out this code out to avoid it being done later however I decided it was not a good idea because:
- The current abstraction would not work because each pin now has its own state, so there would need to be a struct for each pin and a struct for the top level multi progress bar
- Abstracting out new code when it might change to invalidate said abstraction would result in wasted work
- I don't know what use case the code would be reused for to know what the correct abstraction would be

piegamesde:
> I don't know what use case the code would be reused for to know what the correct abstraction would be

I added that abstraction to add a "verify" command, which checks all the pins. The same could also be done with a flag of course, but this wouldn't change the fact that the status messages of the pins would be different.

About the overall design of the channels, is there a way to process the Nix log messages right within the `nix` module and only send high-level status updates instead? The mid term goal would be to switch away from using `nix-prefetch` anyways so would be good to keep that in mind for the API already

Coca162:
> About the overall design of the channels, is there a way to process the Nix log messages right within the `nix` module and only send high-level status updates instead?

Yeah I can just turn it into a new enum instead and extract everything from the logs I want at the sender (I think I might have had that idea shortly after making the pr and forgot :v)

> I added that abstraction to add a "verify" command, which checks all the pins. The same could also be done with a flag of course, but this wouldn't change the fact that the status messages of the pins would be different.

Make sense, I'll see about breaking off this code into a new file (`update.rs`? name suggestion would be appreciated) and making some structs to fulfill that need.

piegamesde:
I'd go for `progress.rs` if the code is primarily focused on progress tracking and the updating business logic is elsewhere.

Beware though, currently the library/cli split of the Rust crate is a bit weird, if you make a new module it won't be a submodule from `cli.rs` but from libnpins itself (i.e. require importing and stuff). It's a bit unfortunate, and if you can find an easy way to fix that go for it, otherwise it's fine just something to keep in mind

Coca162:
> if you make a new module it won't be a submodule from `cli.rs` but from libnpins itself

Turns out you can just `mod progress;` it into `cli.rs` and rust won't bat a eye.

Not sure if its abstracted out enough for a `verify` command but this should be enough to work with later :3

piegamesde:
Kinda agreed. Ideally, npins wouldn't depend on Nix at all, and do everything through snix as a library.

In the meantime, what about downloading the URLs to a temporary location first (so that we can track the progress conventionally) and then ingesting then into the Nix store via a `file://` URI? Would that work? Would the progress bar be worth the double-copy required for this?

PR #158, opened 2025-07-30
Add timestamps to updates for GitHub repositories

xhalo32:
A simple feature addition: fetch the timestamp of the commit during `npins update`. I wanted this because `nix flake update` had it.

If something's not written perfectly, please let me know and I can try to do it better.

> Note: I haven't used my fork for extensive testing yet. I will fix any bugs I find in my fork.

## Features

- Currently supports GitHub.
- After update, displays that the commit timestamp has changed.
- Timestamp is handled as plain text (to avoid adding an extra dependency)
- Displays N/A if timestamp is not available.

## Future work

Add API calls for other providers in `timestamp_url`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ url = { version = "^2.5", features = [ "serde" ] }
 anyhow = "^1.0"
 tokio = { version = "^1.0", features = ["macros", "rt-multi-thread", "process"] }
 log = "^0.4"
-reqwest = { version = "^0.12.0", features = [ "rustls-tls-native-roots" ], default-features = false }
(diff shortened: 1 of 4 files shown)


piegamesde:
I'm not super fan of adding more GitHub-only features, and even with Gitlab and Forgejo support I'd still prefer a git-native solution.

> I wanted this because nix flake update had it.

*Why* did you want this? What makes that feature useful for flakes? Would simply storing a "last updated" timestamp be equally helpful?

acid-bong:
> I'd still prefer a git-native solution

that'll be the latest commit's timestamp

unless (which is the most common occasion) the input is registered as a forge-specific tarball and npins have to rely on that forge's API

Issue #168, opened 2025-09-09
The HTTP Lockable Tarball is incorrectly implemented


Tom-Hubrecht:
When trying to use the lockable tarball mechanism on https://channels.nixos.org/nixos-unstable-small/nixexprs.tar.xz for example, the locked version is not detected. This is due to the fact that by default `reqwest` follows redirects, and that this url gives the link header in a 302 response.

Hence, `reqwest` will follow the 302 and get a different set of headers in the final 200 response that does not contain the link header.

The fix should be to set `redirect(Policy::none())` while building the client used to fetch the tarball information.

piegamesde:
Is that fix sufficient? Shouldn't it be necessary for a full fix to manually follow the redirect chain to check if *any* of the responses contains the header?

Tom-Hubrecht:
> Is that fix sufficient? Shouldn't it be necessary for a full fix to manually follow the redirect chain to check if _any_ of the responses contains the header?

Indeed, that is true

piegamesde:
I tried to implement this properly, and then found out that our http client simply does not support writing an elegant implementation, then found out that *no* readily available http client library supports this, and then found out that this makes sense because it turns out that the protocol is in fundamental violation of HTTP works. The current implementation is the most correct interpretation of the spec, but as seen here that makes it pretty useless. I will proceed with removing http lockable tarball support and replacing it with some explicit user configuration to serve the same purpose.

https://git.lix.systems/lix-project/lix/issues/360

https://git.lix.systems/lix-project/lix/issues/358#issuecomment-3935

PR #169, opened 2025-09-10
tarball: Look for a link header in the whole redirection chain

Tom-Hubrecht:


piegamesde:
Honestly, I hate this, and I'm slightly leaning towards ripping out the entire "lockable URLs" dumpster fire of a protocol again altogether

piegamesde:
https://github.com/andir/npins/issues/168#issuecomment-4275674369

Thanks for your efforts, however the conclusion from discussions with several people at LixCon is that we should just remove this feature again and replace it with something more intentional and less broken altogether.

PR #173, opened 2025-09-24
feat: use NixHash as internal hash type

KFearsoff:
Instead of parsing back and forth from String in many places, use NixHash as the internal data representation, and parse outputs returned from various tools early on.

This depends on a patch to `nix-compat`.


KFearsoff:
Okay, this is a very silly hack that I'm not proud of, but because `Pin::Tarball`'s `hashes` is expected to be a struct that is `flatten`'ed, setting it to `NixHash` (which is a string) results in deserialization failure. So I added a simple wrapper struct.

I feel like there should be a better solution, but for now I'm drawing blanks.

src/nix.rs
```diff
@@ -41,9 +36,16 @@ pub async fn nix_prefetch_tarball(url: impl AsRef<str>) -> Result<String> {
...
-        let stdout = String::from_utf8_lossy(&output.stdout);
-        log::debug!("Got hash: {}", stdout);
-        hash_to_sri(&stdout.trim(), "sha256")
+        // try to parse the returned hash.
+        let hash_str = std::str::from_utf8(&output.stdout)
+            .with_context(|| "nix-prefetch-url sent invalid utf8")?
+            // Trim, otherwise the call to `from_nix_nixbase32` will fail due to newline
+            .trim();
+        // FIXME: nix-compat expects nixbase32-encoded string in format "alg:digest", but
+        // `nix-prefetch-url` gives digest only. Gross way to do this is to specify alg manually
```
piegamesde, line 45: Do we know that nix-prefetch-url always returns the sha256 hash?

src/lib.rs
```diff
@@ -333,14 +334,9 @@ impl diff::Diff for GenericVersion {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
-pub struct GenericHash {
-    pub hash: String,
-}
-
-impl diff::Diff for GenericHash {
+impl diff::Diff for NixHash {
```
piegamesde, line 337: This sounds wrong, because `NixHash` is just a fancy string type and not a keyed container type

src/tarball.rs
```diff
@@ -39,10 +39,29 @@ impl diff::Diff for LockedTarball {
     }
 }
 
+// HACK: this is silly, but we need this because we `serde::flatten` our pins, so we need a struct
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+pub struct HashWrapper {
```
piegamesde, line 44: I don't really understand this type. What is the difference to the previous `GenericHash` type? This change reads like going full circle to me

piegamesde (changes requested):
Thank you for picking this up

src/tarball.rs
KFearsoff: I mean, yeah, pretty much :/

I could've sworn that using `NixHash` directly causes a serde failure due to `#[serde(flatten)]`, but I've just tried locally and can't reproduce? Idk what I've stumbled into before, but yeah, I'm pushing a good version in a sec.

KFearsoff:
Finally, freedom!

Had to do very ugly things to serde to comply. Ugh. But now no wrapper types.

piegamesde (approved):

PR #176, opened 2025-11-11
Explain why use npins in readme

jappeace:
I was just confused why a seperate project was made, so I did a little digging and found that comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,9 @@ Simple and convenient dependency pinning for Nix
 
 ## About
 
-`npins` is a simple tool for handling different types of dependencies in a Nix project. It is inspired by and comparable to [Niv](https://github.com/nmattia/niv).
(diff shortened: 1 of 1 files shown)


piegamesde:
Thank you, but I think I'm going to decline this proposal. Niv is a perfectly fine tool for those who prefer its API over npins', and given how mature Niv is I don't see it being in low-maintenance mode a problem for now.

In fact, when npins development started, Niv was kind of unmaintained and on hiatus, but then eventually ended picking up speed again some year later

jappeace:
@leana8959 said niv don't support pip, and doesn't support forgejo well

arthsmn:
also, there's now [lon](https://github.com/nikstur/lon), which seems pretty similar but is [recommended by lanzaboote](https://github.com/nix-community/lanzaboote/pull/504).

PR #179, opened 2025-11-18
`npins add git` forge autodetection

CommentatorForAll:
This pr adds a `git_auto` function for auto-detection of git forges to optimize pinning. 
A new cli flag `--forge` has been added to force a certain forge when using the `add` command

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -23,6 +23,14 @@ fn get_github_api_url() -> String {
         .unwrap_or_else(|_| String::from("https://api.github.com"))
 }
 
+/// Extract the scheme and host of a domain, stripping the rest
+fn strip_url(mut url: Url) -> Url {
+    url.set_query(None);
+    url.set_fragment(None);
+    url.set_path("");
+    url
+}
+
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
 pub struct GitRevision {
     revision: String,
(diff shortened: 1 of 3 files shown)


src/git.rs
```diff
@@ -118,7 +118,61 @@ pub enum Repository {
...
+    ///
+    /// For forges that can also be self-hosted, only the well-known "main" forge will be taken into account (e.g. gitlab.com).
+    /// Only works for HTTP(S) URLs.
+    /// Invalid URLs will equally fall back to `Self::Git`, as we can't know for sure what is invalid without trying it out later.
+    pub async fn git_auto(url: Url) -> Self {
+        let url2 = url.clone();
+        match (url.scheme(), url.domain()) {
+            ("http" | "https", Some("github.com")) => Self::github_from_url(url),
+            ("http" | "https", Some("gitlab.com")) => Self::gitlab_from_url(url),
+            ("http" | "https", Some("forgejo.org")) => Self::forgejo_from_url(url),
```
piegamesde: forgejo.org is not a forgejo instance. But we can use codeberg.org instead

src/git.rs
```diff
@@ -118,7 +118,61 @@ pub enum Repository {
...
+    }
+
+    ///
+    /// Takes in a URL of unknown forge and tries to determine which forge the hoster is
+    /// And then parse the url into the according Repository Variant
+    async fn probe_forge(url: Url) -> Option<Self> {
+        async fn probe(mut test_url: Url, path: &str) -> Result<bool> {
+            test_url.set_path(path);
+            let res = reqwest::get(test_url).await?;
+            Ok(res.status() != 404)
```
piegamesde: Given that both APIs will return JSON, I'd argue that the response should successfully parse as such. In its current form, I'd be worried that some server throwing a 500 or 403 might yield a false positive

PR #186, opened 2025-12-23
Separate out libnpins and create workspace

Coca162:
Doing the minimum amount of work to split the two apart from making sure we aren't duplicating anything like versions or dependencies shared between them. I'd be happy to change the PR to some other layout/structure if preferred over having libnpins in its own folder and keeping the CLI at the root.


piegamesde:
I think we all agree on the API being not stable, and that will be reflected by SemVer minor bumps (which on 0. versions are defined breaking).

piegamesde:
Can we keep `npins` as the import name?

andir:
@Coca162  Please do another rebase and then this should be good.

PR #187, opened 2025-12-26
Fix regression in pin animation

Coca162:
Fixes the cursor's column not being set back to 0 for the last `on_pin_finish`  in `in_progress`. This currently leaves ghost status message at the last pin due to not being wiped previously.

Introduced in [the factoring out out of `Animation`](https://github.com/andir/npins/commit/2fc0caddc3ebb680b33d8c9475eacaf459dc5b53#diff-b2812f19576dd53d0c35b107a322f58a00fce6977f4b5976e1961853982af3ccL760-L765).

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -780,8 +780,7 @@ impl Opts {
             .map(|(name, pin)| async move {
                 animation.on_pin_start(&*name);
                 let diff = Self::update_one(pin, strategy).await?;
-                animation.on_pin_finish(&*name);
-                animation.write(|stderr| write_diff(stderr, name, &diff));
+                animation.on_pin_finish(&*name, |stderr| write_diff(stderr, name, &diff));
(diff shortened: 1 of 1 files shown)


Coca162:
I've been testing it for a bit and I think I found what issue you might've caught, unfortunately I think there needs to be a way to differentiate between adding and removing from in_progress here because the shared function is causing headaches.

Coca162:
Hopefully this should be fine now, will go and test a bit more (I'd record the issue but turns out screen recording on a wayland window manager is insufferable for me!)

Coca162:
Alrighty I think this should now be fine, it much more closely matches the behavior of the previous code so hopefully nothing else arises, here's a recording of the 2 issues and the fixed one:

https://github.com/user-attachments/assets/fbf6676f-3b38-48ed-a609-5a5586ec7cca

PR #188, opened 2025-12-29
default.nix: allow passing pkgs at the top level

Gerg-L:
This allows you to pass a `pkgs` argument at the top level instead of per-pin if you want to.
i also added a impure-eval check in the override logic which should speed stuff up
then i made mayOverride massively faster by not using per character `builtins.match` but rather one `builtins.split`

i ran `nix-build -A meta.tests`

i have some other changes i'd like to make but they're less important. so i'll make them in another PR

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -19,17 +19,6 @@ let
     in
     (if e.success then e.value else { error = fn { }; }) // { __functor = _self: fn; };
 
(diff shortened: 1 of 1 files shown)


andir:
In that case couldn't you have a tiny wrapper around the sources that does inject that instance into all the fetchers?

Something like:
```
{ pkgs, ... }:
let
   wrapNPins = pkgs: sources: builtins.mapAttrs (_: v: v { inherit pkgs; }) sources;
in wrapNPins pkgs (import ./npins)
```

Issue #175, opened 2025-11-10
Should updating a nonexistent pin be an error?


stuebinm:
I just did `npins update missing` (I was in the wrong project), which produces:

~~~
; npins update missing
[WARN ] Specified pin does not exist: missing
[INFO ] Update successful.
~~~

and exit code 0. It does print a warning & luckily I noticed it, but: personally I'd have expected npins to fail if an update specifies no valid pins at all, or at least not print "Update successful" at the end, since it's not like there was anything it could've even tried to update, and such a case is likely to be a user/scripting error.

(if people here agree with changing this, I could write a patch to change this)

leana8959:
I agree, I think it should be an error.

On the other hand, should updating a frozen pin be an error?

PR #191, opened 2026-01-05
cli: fail on updates with no valid pins

stuebinm:
Minimal patch to address https://github.com/andir/npins/issues/175

If no pin is selected, npins should fail an update. Can happen when:
- there are no pins
- all specified pins do not exist
- all pins were frozen

Points for discussion:
- should we fail if all specified pins were frozen (but *do* exist)?
- should we fail only if *all* specified pins don't exist (this patch) or if only *some* do not exist?

Also a question for the maintainers: is simply returning an anyhow::Error the correct approach here? It is consistent with other errors in npins, but I was a little surprised (and initially implemented the message by using `log::error!` instead, so it prints nicely)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -758,6 +758,10 @@ impl Opts {
             selected_pins.len()
         };
 
+        if length == 0 {
+            return Err(anyhow::anyhow!("no valid pin selected for update"));
+        }
+
         let strategy = match (opts.partial, opts.full) {
             (false, false) => UpdateStrategy::Normal,
             (false, true) => UpdateStrategy::Full,


stuebinm:
> The case where there are no-thawed pins available might be the interesting one where we wouldn't want to fail but perhaps log a warning?

I don't much of an opinion here (I have never used frozen pins). However, if it emits a warning, it would just be one more extra warning below a list of "[WARN ] Ignoring frozen pin", so from a usability perspective I'm not sure it'd add all that much / make it more likely to be noticed? But it's not a big change, so, :shrug:

andir (approved):
I think we can accept it as is. I think it is better than silenty doing nothing.

PR #194, opened 2026-01-08
default.nix: adjust comments

ghost:
Adjusts the comments on the default.nix to be less intrusive. Also no swearing.

### DIFF
--- a/src/default.nix
+++ b/src/default.nix
@@ -65,7 +65,9 @@ let
         if pkgs == null then
           {
             inherit (builtins) fetchTarball fetchurl;
-            # For some fucking reason, fetchGit has a different signature than the other builtin fetchers …
+            # Frustratingly, due to flakes and `fetchTree`, `fetchGit`
+            # has a different signature than the other builtin
+            # fetchers
             fetchGit = args: (builtins.fetchGit args).outPath;
           }
         else
@@ -95,7 +97,6 @@ let
               };
           };
 
-      # Dispatch to the correct code path based on the type
       path =
         if spec.type == "Git" then
           mkGitSource fetchers spec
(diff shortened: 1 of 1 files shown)


src/default.nix
```diff
@@ -65,7 +65,7 @@ let
         if pkgs == null then
           {
             inherit (builtins) fetchTarball fetchurl;
-            # For some fucking reason, fetchGit has a different signature than the other builtin fetchers …
+            # `fetchGit` has a different signature than the other builtin fetchers
```
piegamesde:
```suggestion
            # For some frustrating reason (Flakes), fetchGit has a different signature than the other builtin fetchers …
```

src/default.nix
```diff
@@ -95,7 +95,6 @@ let
               };
           };
 
-      # Dispatch to the correct code path based on the type
```
piegamesde: Not sure what's offending about that comment

piegamesde (changes requested):
The swearing can go, but the frustration will stay

src/default.nix
```diff
@@ -231,7 +230,6 @@ mkFunctor (
       if builtins.isPath input then
         # while `readFile` will throw an error anyways if the path doesn't exist,
         # we still need to check beforehand because *our* error can be caught but not the one from the builtin
-        # *piegames sighs*
         if builtins.pathExists input then
```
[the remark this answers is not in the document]
normalcea: Added, this is an even better solution.

src/default.nix
normalcea: I included the reasoning within the comment (kept the frustrating part too because let's take every moment to point out flakes problems).

PR #198, opened 2026-01-29
Update `reqwest` to 0.13.1

### LINKED ISSUE #197: OS certificate store is not used
I am using NixOS behind a proxy and I have the certificates added to security.pki.certificateFiles in my NixOS config. This works and fixes almost all of my certificates related issues. However npins does not appear to be using the certificates from the certificate store:

```
❯ npins init
[INFO ] Welcome to npins!
[INFO ] Writing default.nix
[INFO ] Writing initial lock file with nixpkgs entry (need to fetch latest commit first)
Error: Failed to fetch initial nixpkgs entry

Caused by:
    0: error sending request for url (https://channels.nixos.org/nixpkgs-unstable/nixexprs.tar.xz)
    1: client error (Connect)
    2: invalid peer certificate: UnknownIssuer
```

Coca162:
`reqwest` now uses `rustls-platform-verifier` by default, which should fix #197

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.15.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b7b6141e96a8c160799cc2d5adecd5cbbe5054cb8c7c4af53da0f83bb7ad256"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.37.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Coca162:
Strange how the tarball test only started failing now, quieted

LyndseyDicksonL3Harris:
That fixed the issue. Thank you!

As an aside, `nix-build .` failed because of the snix package:
`fatal: unable to access 'https://git.snix.dev/snix/snix/': SSL certificate OpenSSL verify result: unable to get local issuer certificate (20)`

However, I was able to successfully invoke `nix-shell` and complete the build with `cargo build --bin npins --features clap,crossterm,env_logger`.

LyndseyDicksonL3Harris:
It looks like my company's firewall blocks https://git.snix.dev/snix/snix.git. Annoying but unsurprising.

winterqt:
The same mechanism is being used here (both `rustls-native-certs` and `rustls-platform-verifier` use `openssl-probe` on Linux), so I'm a bit perplexed as to how this was fixed by migrating 🤔

It's not like `openssl-probe` was bumped in this PR or anything... fun.

Coca162:
In hindsight I realized this issue might be due to using a nixos stable which currently has a older version of npins which still had this issue

PR #201, opened 2026-02-04
Add the hash to `dockerTools.pullImage`.

SamirTalwar:
This passes the hash from the container spec through to `dockerTools.pullImage`.

Without this, I get the following error:

```
outPath = «error: fetcher called without `hash`»
```


andir:
Look reasonable, can you also add a test case? :-)

SamirTalwar:
Will do! Tomorrow though.

andir (approved):

PR #203, opened 2026-02-12
Add completions and fish pin completions

### LINKED ISSUE #152: shell completions
it would be nice to get shell completions for npins commands.
considering a full `npins update` action a bit more intensive in terms of rebuilds, i've found myself manually updating pins for my dependencies. as such, i noticed shell completions were currently missing still, while these might help facilitate this process a bit.

Coca162:
Added compile time generated completions alongside partially hand rolled pin completions for fish. While something using the clap rust native completion engine was a option here, it currently does not give us access to the context to be able to handle lockfile mode. 

Some changes to the CLI were done to make `npins remove` consistent with other operations on pins and also to facilitate being able to write fancier pin completion in fish.

pre-commit hook for keeping completions up to date is handled in the most straightforward way I can imagine because I didn't want to add another nix file like with update-readme.

Partially resolves #152, a new issue should be opened for pin completions on bash and zsh

Best viewed commit by commit


piegamesde:
I've seen it in other software where the completions generation are a generic sub-command of the binary, and then you can just do `. $(npins completions zsh)` in your .zshrc or whatever. What about an approach like this?

Also, what's the benefit of having users know when the completions change? Personally I haven't cared about the code of the completions once, as long as it matches whatever the arguments of the binary are

Coca162:
> I've seen it in other software where the completions generation are a generic sub-command of the binary, and then you can just do `. $(npins completions zsh)` in your .zshrc or whatever. What about an approach like this?

You can, but it felt unnecessary to add extra dependencies to the cli just for completions when most users are going to be using nixos which will package it for them anyways. I'm fine with it either way though.

> Also, what's the benefit of having users know when the completions change? Personally I haven't cared about the code of the completions once, as long as it matches whatever the arguments of the binary are

Ah to be clear its not about the users being notified of the change its about a developer knowing to so modified completions can be updated as well. This might be avoided by maybe making the pin completions a patch on top of the auto generated completions instead of the whole thing but that feels like it could be fragile.

piegamesde:
If this adds more dependencies to the CLI then having it in a separate binary might indeed be preferable. 

> Ah to be clear its not about the users being notified of the change its about a developer knowing to so modified completions can be updated as well.

That's what I don't understand, if we generated the completions unconditionally as part of the packaging, then this kind of manual updating would be not necessary, right?

Coca162:
> That's what I don't understand, if we generated the completions unconditionally as part of the packaging, then this kind of manual updating would be not necessary, right?

We cannot generate it unconditionally if you want pin completions, you have to modify the generated completions to add it in, and clap_complete does not have the ability (even with the dynamic completions engine) to handle our use case currently so we need some modification somewhere.

piegamesde:
Aah, I understand the issue now. But also I may need some help with understanding more basic questions here. How does updating the generated code mixed in with manually edited shell code work? For which CLI arguments do we offer dynamic completions, and why only in fish?

Coca162:
>  How does updating the generated code mixed in with manually edited shell code work?

For now, I just go and diff the generated code and the final completions and by hand move over all the changes to the final file.

>  For which CLI arguments do we offer dynamic completions

`show`, `update`, `verify`, `freeze`, `unfreeze`, `remove` all need to be supplied with pins from your lockfile so all of those.

Additionally, `--name` for `add` also completes on existing pins in your lockfile since you can replace pins with the command.

> why only in fish?

It was a pain for me to figure out fish and I am not excited to work on the other shells since I don't use them. I'm hoping that someone who already knows bash/zsh and is motivated enough to get the initial chunk of work done could drive-by contribute it (I'd be happy to maintain it after that).

piegamesde:
> For now, I just go and diff the generated code and the final completions and by hand move over all the changes to the final file.

How easy would it be to extract your manual contributions into a patch? The patch would then fail to apply if things diverge too strongly and require manual fixing. But then at least it would be clear what is auto generated code and what is out edit to it

Coca162:
> How easy would it be to extract your manual contributions into a patch? The patch would then fail to apply if things diverge too strongly and require manual fixing. But then at least it would be clear what is auto generated code and what is out edit to it

I'm not exactly sure which is why I kept it this way for now. I'm also not that familiar with having to fix a patch by hand, I'd usually just handle conflicts in git. I'd be okay with trying it out though.

piegamesde:
My process of updating patches in Nix is to simply use the Git tools and regenerate it. Stuff like apply on the old code, rebase on the new generated code, rediff. But yeah, all of this is unfortunate, to the point where I question if it wouldn't be easier to simply skip dynamic completions. Having *any* completions on most commands is worth the most, the rest is just a bonus IMO

Coca162:
I do find it very convenient and something I'd be happy to maintain despite its burden. I've already had to update it as the CLI has changed since this PR first got released and it wasn't too bad.

Coca162:
I'm very happy with the way this is done now. I found out that you can append them and thusly I don't have to deal with modifying the autogenerate commands. Not sure if this would work for other shells but its incredibly convenient for fish.

piegamesde (approved):

PR #214, opened 2026-04-09
Add file pin type for plain URL downloads

### LINKED ISSUE #163: support arbitrary downloads
Currently, `npins add` supports a number of difference source types:

```
$ npins add --help
Adds a new pin entry

Usage: npins add [OPTIONS] <COMMAND>

Commands:
  channel  Track a Nix channel
  github   Track a GitHub repository
  forgejo  Track a Forgejo repository
  gitlab   Track a GitLab repository
  git      Track a git repository
  pypi     Track a package on PyPi
  tarball  Track a tarball
```

One option potentially missing here seems non-tarball downloads, which seems useful for e.g. NixOS ISO downloads.
To be fair, this may differ from existing source types in the sense it would offer a file (like `pkgs.fetchurl`) rather than a directory.

KiaraGrouwstra:
Implements support for tracking arbitrary file downloads (non-tarball). Uses `pkgs.fetchurl to fetch a single file by URL and hash, without unpacking.

Usage: `npins add file <url>`

closes: #163.

disclaimer i used a coding agent in the creation of this patch.

### DIFF
--- a/libnpins/src/fetchurl.rs
+++ b/libnpins/src/fetchurl.rs
@@ -99,23 +99,4 @@ mod test {
         assert_eq!(value["url"], "https://example.com/file.bin");
         assert!(value["hash"].as_str().unwrap().starts_with("sha256-"));
     }
-
-    #[tokio::test]
-    async fn test_fetchurl_update_and_fetch() {
-        // A small file served from the Nix cache. It is not strictly
-        // immutable (new fields may be added over time), but in practice it
-        // is stable enough to make for a reasonable fetch target in tests.
-        let pin = FetchurlPin {
-            url: "https://cache.nixos.org/nix-cache-info".parse().unwrap(),
-        };
-
-        let version = pin.update(None).await.unwrap();
-        assert_eq!(version, FetchurlVersion {});
-
-        let hashes = pin.fetch(&version).await.unwrap();
-        assert_eq!(
-            hashes.hash,
-            NixHash::from_sri("sha256-LJ3jc651pScWN2NQNERaXNOmrjWsbDBtQMDgZ2R4WJc=").unwrap(),
-        );
-    }
 }
--- a/libnpins/src/fetchurl.rs
+++ b/libnpins/src/fetchurl.rs
@@ -11,36 +11,36 @@ use url::Url;
 use crate::{GenericHash, Updatable, diff, nix};
 
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash)]
-pub struct FilePin {
+pub struct FetchurlPin {
     pub url: Url,
 }
 
-impl diff::Diff for FilePin {
+impl diff::Diff for FetchurlPin {
     fn properties(&self) -> Vec<(String, String)> {
         vec![("url".into(), self.url.to_string())]
     }
 }
 
-/// File pins have no meaningful version; the URL is assumed to be static.
+/// Fetchurl pins have no meaningful version; the URL is assumed to be static.
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
-pub struct FileVersion {}
+pub struct FetchurlVersion {}
 
-impl diff::Diff for FileVersion {
+impl diff::Diff for FetchurlVersion {
     fn properties(&self) -> Vec<(String, String)> {
         vec![]
     }
 }
 
 #[async_trait::async_trait]
-impl Updatable for FilePin {
-    type Version = FileVersion;
+impl Updatable for FetchurlPin {
+    type Version = FetchurlVersion;
     type Hashes = GenericHash;
 
-    async fn update(&self, _old: Option<&FileVersion>) -> Result<FileVersion> {
-        Ok(FileVersion {})
+    async fn update(&self, _old: Option<&FetchurlVersion>) -> Result<FetchurlVersion> {
+        Ok(FetchurlVersion {})
     }
 
-    async fn fetch(&self, _version: &FileVersion) -> Result<Self::Hashes> {
+    async fn fetch(&self, _version: &FetchurlVersion) -> Result<Self::Hashes> {
         let hash = nix::nix_prefetch_url(&self.url).await?;
         Ok(Self::Hashes { hash })
     }
@@ -53,15 +53,15 @@ mod test {
     use nix_compat::nixhash::NixHash;
 
     #[test]
-    fn test_file_pin_roundtrip() {
+    fn test_fetchurl_pin_roundtrip() {
         let pins = NixPins {
             pins: [(
                 "test-file".into(),
-                Pin::File {
-                    input: FilePin {
+                Pin::Fetchurl {
+                    input: FetchurlPin {
                         url: "https://example.com/some-file.iso".parse().unwrap(),
                     },
-                    version: Some(FileVersion {}),
+                    version: Some(FetchurlVersion {}),
                     hashes: Some(GenericHash {
                         hash: NixHash::from_sri(
                             "sha256-K9yBph93OLTNw02Q6e9CYFGrUhvEXnh45vrZqIRWfvQ=",
@@ -81,12 +81,12 @@ mod test {
     }
 
     #[test]
-    fn test_file_pin_serialization_shape() {
-        let pin = Pin::File {
-            input: FilePin {
+    fn test_fetchurl_pin_serialization_shape() {
+        let pin = Pin::Fetchurl {
+            input: FetchurlPin {
                 url: "https://example.com/file.bin".parse().unwrap(),
             },
-            version: Some(FileVersion {}),
+            version: Some(FetchurlVersion {}),
             hashes: Some(GenericHash {
                 hash: NixHash::from_sri("sha256-K9yBph93OLTNw02Q6e9CYFGrUhvEXnh45vrZqIRWfvQ=")
                     .unwrap(),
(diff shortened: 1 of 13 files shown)


fricklerhandwerk:
Ah awesome! Just the other day at the Nixpkgs security tracker I felt this missing, because we needed that one file as input and couldn't express it: https://github.com/NixOS/nix-security-tracker/pull/951/files

libnpins/src/file.rs
```diff
@@ -0,0 +1,118 @@
...
+        assert_eq!(value["type"], "File");
+        assert_eq!(value["url"], "https://example.com/file.bin");
+        assert!(value["hash"].as_str().unwrap().starts_with("sha256-"));
+    }
+
+    #[tokio::test]
+    async fn test_file_update_and_fetch() {
+        // Use a small, stable file from the Nix cache
+        let pin = FilePin {
+            url: "https://raw.githubusercontent.com/NixOS/nixpkgs/master/README.md"
```
samueldr: - Not stable
 - Not from the Nix cache

libnpins/src/file.rs
```diff
@@ -0,0 +1,118 @@
...
+                .parse()
+                .unwrap(),
+        };
+
+        let version = pin.update(None).await.unwrap();
+        assert_eq!(version, FileVersion {});
+
+        let hashes = pin.fetch(&version).await.unwrap();
+        // Just verify we got a sha256 hash back
+        assert_eq!(hashes.hash.algo(), nix_compat::nixhash::HashAlgo::Sha256);
```
samueldr: If this is intended to be ran against a stable resource, why not check the hash that was returned?

libnpins/src/nix.rs
```diff
@@ -10,6 +10,41 @@ pub struct PrefetchInfo {
...
+            .arg("--name")
+            .arg("source")
+            .arg("--type")
+            .arg("sha256")
+            .arg(url)
+            .output()
+            .await
+            .with_context(|| format!("Failed to spawn nix-prefetch-url for {}", url))?;
+
+        if !output.status.success() {
```
samueldr, line 29:
```suggestion
        // FIXME: handle errors and pipe stderr through
        if !output.status.success() {
```

This pattern was clearly copied over from the other functions, but stripped of their comments.

Is there any reason for that?

libnpins/src/nix.rs
```diff
@@ -10,6 +10,41 @@ pub struct PrefetchInfo {
...
+                "failed to prefetch url: {}\n{}",
+                url,
+                String::from_utf8_lossy(&output.stderr)
+            )));
+        }
+
+        let hash_str = std::str::from_utf8(&output.stdout)
+            .with_context(|| "nix-prefetch-url sent invalid utf8")?
+            .trim();
+        let annotated_nix32_hash = format!("sha256:{hash_str}");
```
samueldr, line 44:
```suggestion
        // FIXME: nix-compat expects nixbase32-encoded string in format "alg:digest", but
        // `nix-prefetch-url` gives digest only. Gross way to do this is to specify alg manually
        let annotated_nix32_hash = format!("sha256:{hash_str}");
```

Same.

libnpins/src/nix.rs
```diff
@@ -10,6 +10,41 @@ pub struct PrefetchInfo {
...
+pub async fn nix_prefetch_url(url: impl AsRef<str>) -> Result<NixHash> {
+    let url = url.as_ref();
+    let result = async {
+        log::debug!(
+            "Executing `nix-prefetch-url --name source --type sha256 {}`",
+            url
+        );
+        let output = tokio::process::Command::new("nix-prefetch-url")
+            .arg("--name")
+            .arg("source")
```
samueldr: The tarball fetcher has a comment explaining this rationale:

> use the same symbolic store path name as `builtins.fetchTarball` to avoid downloading the source twice

What about `builtins.fetchurl` and `mkFileSource`?

libnpins/src/file.rs
```diff
@@ -0,0 +1,118 @@
+//! Pin a plain file URL source
+//!
+//! Uses `builtins.fetchurl` (or `pkgs.fetchurl` when nixpkgs is provided)
+//! to fetch a single file by URL and hash.
```
samueldr: Is this correct for this module?

 - Is the distinction that `pkgs.fetchurl` used important?
 - Is the actual Nix construct even *used* within this Rust module?

tests/default.nix
```diff
@@ -151,6 +153,17 @@ let
       sha256 = hash;
     };
 
+  mkFileSource =
```
andir: We should probably bump the version of the lock file, so people are asked to run the migration.

libnpins/src/fetchurl.rs
```diff
@@ -0,0 +1,118 @@
+//! Pin a plain file URL source
```
samueldr, line 1: (See the main review comment.)

samueldr (changes requested):
I will not comment on the “coding agent” situation, as this is not my project.

Though since I care about `npins`, I will make sure it doesn't sloppify. Which is why I'm reviewing this.

* * *

Some other thoughts on top of the review comments.

I'm not sure that `file` is a good description of what this is intended to do. Neither is "plain file URL".

What this is, I think, is closer to a hardcoded input, a pre-defined input, an (expected to be) immutable input. I don't have a precise suggestion though.

A `file` input to me should be left for... A file.

And an URL input *probably* shouldn't exist as such, and instead be qualified in some ways, e.g. if support for pinning a followed `Location` header is added, such a name could be `followed-url`.

Thinking about it just a tad more, it might make sense to instead call this `fetchurl` since it, well, is intended to be a thin wrapper around `fetchurl` semantics.

fricklerhandwerk:
@Samueldr but isn't a URL referring to a resource that will amount to a file on disk for our purposes? It doesn't seem inconsistent with the rest of the naming either, which can somewhat cleanly be mapped to the object we obtain and then process. The platform-specific ones can also be understood as e.g. "`github` means the thing that GitHub exposes, a repo or whatnot". After all, we always refer to resources as URLs, for all sources, and they always resolve to file( system object)s eventually. So why not name them by what distinguishes them?

samueldr:
> So why not name them by what distinguishes them?

That's why in the end I thought about:

> Thinking about it just a tad more, it might make sense to instead call this `fetchurl` since it, well, is intended to be a thin wrapper around `fetchurl` semantics.

What distinguishes this feature is not that this results in a “file” or uses an “URL” as an input, but that it uses *the semantics of* `fetchurl`.

Using your example, the `github` fetch doesn't result in a github, and doesn't take a github as an input.

Another example, while `tarball` effectively uses a tarball, it takes an URL as an input, but does not result in a tarball: it follows the `fetchTarball` semantics.

Using `file` lacks precision. I also had dropped some explanations to prevent unneeded side-discussions, but that that would clobber the `file` identifier for adding support for file-based pins. While it might sound super redundant, since you can just use the file right there in Nix, I think it would make sense in some form to have `file` work as an `npins` input at some point. Either as an affordance so users can replace a pin more easily, or simply so all your inputs use the same semantics, so you don't need to care where it is in the project, just that it is an input.

libnpins/src/fetchurl.rs
```diff
@@ -0,0 +1,121 @@
...
+        assert_eq!(value["type"], "Fetchurl");
+        assert_eq!(value["url"], "https://example.com/file.bin");
+        assert!(value["hash"].as_str().unwrap().starts_with("sha256-"));
+    }
+
+    #[tokio::test]
+    async fn test_fetchurl_update_and_fetch() {
+        // A small, immutable file served from the Nix cache. Its contents
+        // (store dir, priority, ...) are part of the cache's public contract
+        // and are not expected to change.
```
samueldr: I'm not sure this is a claim that is *absolutely true* for `Priority`, and nothing prevents this to expose more properties later down the line. So while it's not expected to change, it still could.

So, two main thoughts here:

 - Is this unit test useful when we have the integration test basically doing the same thing?
 - If it is, why not use the same resource, which that one *should* actually be static and not expected to change?

libnpins/src/fetchurl.rs
```diff
@@ -0,0 +1,121 @@
...
+                    .unwrap(),
+            }),
+            frozen: Frozen::default(),
+        };
+
+        let value = serde_json::to_value(&pin).unwrap();
+        assert_eq!(value["type"], "Fetchurl");
+        assert_eq!(value["url"], "https://example.com/file.bin");
+        assert!(value["hash"].as_str().unwrap().starts_with("sha256-"));
+    }
```
samueldr, line 101: I didn't comment on these earlier because I needed to ask someone who was more rust-brained to see if my thoughts were correct.

My thoughts were that this round-trip through the rust type system isn't valuable as a unit test, as the properties of the serialization library and the type system are being tested, really, and not any feature from this module.

*Especially* given that we have the integration test that exercises the deserialization aspect.

But don't put too much weight on my opinion alone for these. Instead defer to the maintainers.

src/main.rs
```diff
@@ -386,6 +399,12 @@ pub enum AddCommands {
     /// URL which supports flakes "Lockable HTTP Tarball" API.
     #[command(name = "tarball")]
     Tarball(TarballAddOpts),
+    /// Track a plain file download
+    ///
+    /// Tracks a plain file URL. Unlike tarball, this fetches a single file
+    /// without unpacking. Useful for e.g. ISO images or individual files.
```
samueldr: Still mainly thinking about *precision* here, not trying to bikeshed anything. Feel free to reword differently (and rewrap).

The main concern is ensuring the help text properly conveys that this is meant for unchanging URLs.

There really isn't *track[ing]*, at least for what I consider tracking: the resource is used as-is, no fancy `Location:` follows, you get what you asked for.

```suggestion
    /// Pin a file through an URL
    ///
    /// Pins a plain file through an URL. There is no additional processing
    /// for the URL, redirects are followed at download time but not tracked.
    /// There is no further processing of the downloaded artifact.
    /// Unlike e.g. tarball, this fetches a single file without unpacking.
    /// Useful for unchanging static assets e.g. ISO images or any other
    /// individual files.
```

samueldr (changes requested):
One (not that) "major" change requested. The comments around the tests really are more comments than requests.

Make sure any other wording in the PR that would be affected by the suggestion at the subcommand's help text is also edited accordingly.

KiaraGrouwstra:
@samueldr thanks again for all your feedback - i've incorporated them now.

KiaraGrouwstra:
@andir I think the CI failure is expected due to the version bump

piegamesde:
#224, #223

PR #215, opened 2026-04-10
accept short git revision hashes for `--at` flag

### LINKED ISSUE #166: accept short hashes for `--at`
currently, if one passes a short revision hash to the `--at` flag of `npins add`, one is met with the error [`'{revision}' is not a valid git revision (sha1 hash)`](https://github.com/andir/npins/blob/afa9fe50cb0bff9ba7e9f7796892f71722b2180d/src/git.rs#L34).
given the likes of hydra would communicate short hashes (e.g. 'Input changes' column at their [jobset page](https://hydra.nixos.org/jobset/nixpkgs/trunk)), it would be cool if npins accepted these as well.

whether this is technically feasible / practical tho, i don't know.

edit: [`git rev-parse <revision>`](https://git-scm.com/docs/git-rev-parse) seems relevant

KiaraGrouwstra:
Allow 7-40 character hex strings as git revisions, not just full 40-character SHA-1 hashes. This makes it possible to use short hashes.

implements the approach suggested at https://github.com/andir/npins/issues/166#issuecomment-3364631894, which we may improve on in the future.

closes #166.

dsisclaimer i used a coding agent in the creation of this patch.

### DIFF
--- a/libnpins/src/git.rs
+++ b/libnpins/src/git.rs
@@ -41,8 +41,10 @@ pub struct GitRevision {
 
 impl GitRevision {
     pub fn new(revision: String) -> Result<Self> {
-        if !revision.chars().all(|c| c.is_ascii_hexdigit()) || revision.len() != 40 {
-            anyhow::bail!("'{revision}' is not a valid git revision (sha1 hash)");
+        if !revision.chars().all(|c| c.is_ascii_hexdigit()) || !(7..=40).contains(&revision.len()) {
+            anyhow::bail!(
+                "'{revision}' is not a valid git revision (expected 7-40 hex characters)"
+            );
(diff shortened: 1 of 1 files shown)


piegamesde:
This lacks end-to-end tests, so has this been tested for functionality? According to https://github.com/andir/npins/issues/166#issuecomment-3652131194, this approach has been deemed unworkable, as much as I'd like to be convinced of the opposite

KiaraGrouwstra:
whoops, i suppose getting to the point of obtaining an `.outPath` didn't tell much after all - indeed, this still got stuck on short hashes in `sources.json` getting rejected on fetch.

let's close then, until things change.

PR #216, opened 2026-04-10
Show forge compare URLs on update

### LINKED ISSUE #177: [feature request] view change history when updating pins
It would be nice to add a flag to view the git log of the upstream changes when doing `npins update` and not just see the rev changing :D

Implementation wise, we could probably pipe that to a pager to make it easier to view.

KiaraGrouwstra:
When a git-based pin changes, print a link to the forge's compare view (e.g. GitHub/Forgejo `/compare/old...new`, GitLab `/-/compare`) so users can quickly review upstream changes.

closes #177.

dsisclaimer i used a coding agent in the creation of this patch.


piegamesde:
#224

PR #217, opened 2026-04-10
feat: add `--arch` flag to `add container`

felixscheinost:
Multi-arch images are very common.
It is also common to don’t have tags suffixed with the arch. That means that currently there is no way to pin those images.


libnpins/src/default.nix
```diff
@@ -214,16 +214,19 @@ let
...
-        hash = hash;
-      };
+      pkgs.dockerTools.pullImage (
+        {
+          imageName = image_name;
+          imageDigest = image_digest;
+          finalImageTag = image_tag;
+          hash = hash;
+        }
+        // (if args ? "arch" then { arch = args.arch; } else { })
```
felixscheinost: We might also need to handle `null`?

But `null` shouldn't be serialized at all using the serde skip_serializing_if.

samueldr: If that's desirable, the following pattern can be used instead:

```suggestion
        // (if args.arch or null != null then { arch = args.arch; } else { })
```

felixscheinost:
Thanks for the feedback and sorry for the delay!

I integrated your suggestion and rebased on master.

It seems there is a inconsistency in the README on master?

When I run the `update-readme` pre-commit hook on master, I also get the same diff.

piegamesde:
The readme diff smells like clap changed its help format in an update and dependabot doesn't run any pre-commit hooks. You can simply pull it in a separate commit and leave it in here

Coca162 (approved):
For the README, I'm not sure exactly what happened since clap versions didn't update but its fine either way.

Don't have really much to test this with because I don't use docker at all but since this is basically passing through a argument I'll assume its fine.

A addition to the CHANGELOG.md would be nice but I'll merge this later either way (already going to go through all new commits to make sure we don't miss mentioning something in the next release).

PR #220, opened 2026-04-18
CI: Also try macOS

samueldr:
Both `x86_64` and `aarch64`.

This either is going to be a pain and will be commented soon enough after introduction, at which point we'll know. Or it'll be useful to keep tabs on the build status for macOS.

CC @andir; I'm guessing since [it's only been a couple of months](https://github.com/andir/npins/commit/b6f34245408412ec7a3ef022d583a08ed937868d), that you're still (un)fortunate enough to be using macOS at work, right? That might be useful for you then.

This might also help with getting PRs like #80 over the hump, in the future, if ever needed.

~~(Initially opened as draft, as GitHub's macOS is still running on my test branch, I'm impatient, and foolishly confident.)~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,13 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4.1.3
       - uses: samueldr/lix-gha-installer-action@8dc19fbd6451fa106a68ecb2dafeeeb90dff3a29 # latest @ 2025-02-28
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
+          - macos-26-intel
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -28,4 +35,5 @@ jobs:
       - name: Run smoke test
         run: nix-shell --run "bash ./.github/workflows/smoke-test.sh"
       - name: Run integration tests
+        if: ${{ runner.os == 'Linux' }}
```
piegamesde, line 38: why only run them on Linux? Does this mean we only test the build on darwin?

samueldr: Oops, I should have maybe hinted at that in the PR body, [in the commit message I did elaborate](https://github.com/andir/npins/pull/220/changes/898370c9ea0c25b955148310b417dce346d1047a).

The TLDR: integration suite doesn't seem to be able to run *as-is* on the hell platform.

samueldr: (And I don't have *the hell platform* handy to look at to see why and how that could be solved.)

andir:
Why do we bother with that OS? I still have that machine.. can try one of these days but motivation for the cursed hardware is near zero...

winterqt:
If I have to be "that person [who cares about macOS]" for npins, I will be. I use macOS for personal use, for better or worse.

andir:
Then the testing burden is on you :-) I've just resigned from my Mac-Job, and I'll have only Linux machines going forward.

PR #221, opened 2026-04-19
Reduce boilerplate around verbosely-logged commands, and unify

samueldr:
Drafted since there are FIXMEs regarding idiomatic Rust stuff, some little work items, and because of the likely clash with what @piegamesde was working on yesterday.

I'm looking for comments regarding if this is desirable first, and then some comments regarding my mishandling of Rust.

I'm also open for opinions or thoughts regarding the "API design" used here. I don't know if there's any way to make this much more handy to use with the way a `tokio::process::Command` works, or even idiomatic Rust..

* * *

The rationale for these changes is, first, how awful it is to go through code that half-duplicates the approximate same ways to print the command that is being ran, but with small details like adding spaces to the arguments conditionally so the command is pretty printed nicely.

And then seeing that at least `nix_eval_pin` had drifted from the actual command being ran.

So instead, let's use a function that *formats* a `tokio::process::Command`, such that it is impossible for it to be incorrect\*

\* It will be "incorrect" for multi-line arguments in its initial draft, by design.

Additionally, this is going through `shlex` for escaping the printed arguments. This should *help* ensuring the commands as-printed can be ran by the ed-user. (Note that I won't claim any safety guarantee, but `shlex` [makes some claims](https://docs.rs/shlex/latest/shlex/).) For this usage, this is for convenience.

Disclaimer: I write my own code. All mistakes are handcrafted with love.

### DIFF
--- a/libnpins/src/git.rs
+++ b/libnpins/src/git.rs
@@ -14,7 +14,9 @@ use serde::{Deserialize, Serialize};
 use tokio::process::Command;
 use url::Url;
 
-use crate::{GenericVersion, Updatable, check_git_url, diff, get_and_deserialize, nix};
+use crate::{
+    GenericVersion, Updatable, check_git_url, diff, format_command, get_and_deserialize, nix,
+};
 
 fn get_github_url() -> String {
     std::env::var("NPINS_GITHUB_HOST").unwrap_or_else(|_| String::from("https://github.com"))
@@ -683,13 +685,17 @@ impl RemoteInfo {
 /// Convenience wrapper around calling `git ls-remote`
 async fn fetch_remote(url: &str, args: &[&str]) -> Result<Vec<RemoteInfo>> {
     let result = async {
-        log::debug!("Executing `git ls-remote {}`", args.join(" "));
-        let process = Command::new("git")
+        let mut command = Command::new("git");
+        command
             // Disable any interactive login attempts, failing gracefully instead
             .env("GIT_TERMINAL_PROMPT", "0")
             .env("GIT_SSH_COMMAND", "ssh -o StrictHostKeyChecking=yes")
             .arg("ls-remote")
-            .args(args)
+            .args(args);
+
+        log::debug!("Executing: {}", format_command(&command)?);
+
+        let process = command
             .output()
             .await
             .context("Failed waiting for git ls-remote subprocess")?;
--- a/libnpins/src/lib.rs
+++ b/libnpins/src/lib.rs
@@ -393,6 +393,82 @@ impl diff::Diff for GenericUrlHashes {
     }
 }
 
+fn elide_to_first_line(input_string: &str) -> String {
+    let mut lines = input_string.lines();
+    let value = lines.next().unwrap_or("(unknown)");
+
+    // Add the ellipsis at the end of the string if needed.
+    match lines.next() {
+        // Only one line.
+        None => value.to_string(),
+        // Multi-line
+        Some(_) => {
+            format!("{}…", value)
+        },
+    }
+}
+
+/// Formats a command in a shell-safe manner.
+///
+/// NOTE: Multi-line components will be elided to their first line!
+pub(crate) fn format_command(tokio_cmd: &tokio::process::Command) -> anyhow::Result<String> {
+    // `tokio::process`'s `Command` doesn't allow introspecting, so let's ignore that
+    // and use it as a `std::process::Command`.
+    let cmd = tokio_cmd.as_std();
+
+    let mut command_parts: Vec<String> = Vec::new();
+
+    // Format and escape environment
+    let envs: Vec<String> =
+        cmd.get_envs()
+            .map(|(name, value)| {
+                let value = elide_to_first_line(
+                value
+                .expect("Environment variable for command should have an environment variable name")
+                .to_str()
+                .unwrap_or("(unknown)")
+            );
+
+                // Escape the value only.
+                let value = shlex::try_quote(&value).unwrap_or("(invalid)".into());
+
+                // The escaping would produce 'NAME=VA LUE', which will not produce a strings that can
+                // be used for passing an environment variables to a command.
+                let name = name.to_str().unwrap_or("(unknown)");
+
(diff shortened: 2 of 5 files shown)


libnpins/src/git.rs
```diff
@@ -683,13 +685,18 @@ impl RemoteInfo {
...
+        // FIXME: idiomatic way for this.
+        let command = command
             // Disable any interactive login attempts, failing gracefully instead
             .env("GIT_TERMINAL_PROMPT", "0")
             .env("GIT_SSH_COMMAND", "ssh -o StrictHostKeyChecking=yes")
             .arg("ls-remote")
-            .args(args)
+            .args(args);
+
+        log::debug!("Executing: {}", format_command(command)?);
```
Coca162:
```suggestion
        command
            // Disable any interactive login attempts, failing gracefully instead
            .env("GIT_TERMINAL_PROMPT", "0")
            .env("GIT_SSH_COMMAND", "ssh -o StrictHostKeyChecking=yes")
            .arg("ls-remote")
            .args(args);

        log::debug!("Executing: {}", format_command(&command)?);
```

You don't need to shadow `command` as the methods for `Command` are on `&mut self`. This can be done everywhere else as well.

libnpins/src/lib.rs
```diff
@@ -393,6 +393,82 @@ impl diff::Diff for GenericUrlHashes {
...
+        Some(_) => {
+            format!("{}…", value)
+        },
+    }
+}
+
+/// Formats a command in a shell-safe manner.
+///
+/// NOTE: Multi-line components will be elided to their first line!
+pub fn format_command(tokio_cmd: &tokio::process::Command) -> anyhow::Result<String> {
```
Coca162:
```suggestion
pub(crate) fn format_command(tokio_cmd: &tokio::process::Command) -> anyhow::Result<String> {
```
Since this doesn't belong in `libnpins` public API

libnpins/src/nix.rs
```diff
@@ -13,17 +13,20 @@ pub struct PrefetchInfo {
...
+        let command = output
             .arg("--unpack") // force calculation of the unpacked NAR hash
             .arg("--name")
             .arg("source") // use the same symbolic store path name as `builtins.fetchTarball` to avoid downloading the source twice
             .arg("--type")
             .arg("sha256")
-            .arg(url)
+            .arg(url);
+
+        log::debug!("Executing: {}", format_command(command)?);
```
Coca162:
```suggestion
        let mut command = tokio::process::Command::new("nix-prefetch-url");
        command
            .arg("--unpack") // force calculation of the unpacked NAR hash
            .arg("--name")
            .arg("source") // use the same symbolic store path name as `builtins.fetchTarball` to avoid downloading the source twice
            .arg("--type")
            .arg("sha256")
            .arg(url);

        log::debug!("Executing: {}", format_command(&command)?);
```

Coca162 (changes requested):
This could be a unit struct or trait extension but that'd probably be unintuitive and cursed. 

It just being a function isn't really all that bad for our uses.

libnpins/src/git.rs
samueldr: I think I (now) got that right, do tell if I missed something.

libnpins/src/nix.rs
```diff
@@ -149,26 +142,23 @@ pub async fn nix_prefetch_docker(
...
-    let output = match image_digest {
-        Some(x) => output.arg("--image-digest").arg(x),
-        None => output,
+
+    match image_digest {
+        Some(value) => {
+            command.arg("--image-digest").arg(value);
+        },
+        None => (),
     };
```
Coca162: This can be turned into a `if let`

libnpins/src/git.rs
```diff
@@ -683,13 +685,17 @@ impl RemoteInfo {
 /// Convenience wrapper around calling `git ls-remote`
 async fn fetch_remote(url: &str, args: &[&str]) -> Result<Vec<RemoteInfo>> {
     let result = async {
-        log::debug!("Executing `git ls-remote {}`", args.join(" "));
-        let process = Command::new("git")
+        let mut command = Command::new("git");
+        command
```
Coca162, line 689: very nitty: a new line between here should be added here or removed from the other uses to keep it consistent (imo I think it looks better without a new line)

libnpins/src/lib.rs
```diff
@@ -393,6 +393,82 @@ impl diff::Diff for GenericUrlHashes {
...
+    let mut command_parts: Vec<String> = Vec::new();
+
+    // Format and escape environment
+    let envs: Vec<String> =
+        cmd.get_envs()
+            .map(|(name, value)| {
+                let value = elide_to_first_line(
+                value
+                .expect("Environment variable for command should have an environment variable name")
+                .to_str()
```
Coca162, line 428: These checked str castings _could_ instead be `.to_string_lossy()` but displaying output that is slightly incorrect in a obfuscated manner is probably not ideal for a edge case like this. The only other way I can imagine dealing with it would be by writing it to a vec or to a formatter directly (with [`fmt::from_fn`](https://doc.rust-lang.org/std/fmt/fn.from_fn.html)).

This is something I think is fine to deal with later and not in this PR however.

Coca162 (changes requested):
Looks good, just some minor nits

libnpins/src/nix.rs
samueldr: Aaah right, that's the idiom I had in mind but couldn't remember.

libnpins/src/git.rs
samueldr: Agreed.

Coca162 (approved):
Great work, happy to merge unless there is anything @piegamesde has to say (following up from the comment that they had clashing work)

piegamesde:
I have a refactoring where I moved all pins into their own module locally, but honestly go for it and I'll figure out the rebase. I can give a second review tomorrow if you want to but I trust the change to be good enough either way

Coca162:
> I trust the change to be good enough either way

same

PR #231, opened 2026-05-05
feat: add git-lfs support

javalsai:
Just adds a field called `lfs` to git pins defaulting to `false` that is forwarded to `builtins.fetchGit`. `fetchGit` has had the `lfs` argument since `2.27`.

My concern is if this should be accompanied by a bump in the pins sources version?

It's also important to consider that the nar hash depends on whether lfs is enabled or not, even for the same revisions.

I made sure to run all commit and push hooks, code is formatted and readme updated. Tested and works for my personal use, this is ready to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,9 @@ Options:
       --submodules
           Also fetch submodules
 
+      --lfs
+          Also fetch LFS references
+
   -h, --help
           Print help (see a summary with '-h')
 ```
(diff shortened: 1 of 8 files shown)


javalsai:
I'm missing some stuff not required to work but that should be there, my next push will have them as well as pass the tests.

javalsai:
This won't pass tests either, as far as I could investigate, the nix shell environment uses lix instead of nix, which apparently doesn't have the `lfs` option yet.

This means that this will break in lix environments, but still works for nix >= 2.27.

piegamesde:
I appreciate your effort, however Lix doesn't support LFS and it doesn't seem like it realistically will anytime soon. Given that Lix is the preferred implementation to use with npins (we use it in CI, I'm a Lix dev, and we've had CppNix specific failures/regressions in the past), I would rather not see this merged unfortunately.

javalsai:
Makes sense. I'll close this for now then but I'll *try* to keep my fork up with to date as I need this.

I'll see if I can do anything to get work for this on lix going, and if it ever comes to lix and its nixpkgs version we can consider reopening this?

PR #232, opened 2026-05-07
Random tidbits and reducing amount of async code

Coca162:
After having used the `current_thread` runtime for a while I'm pretty confident we don't have any issues with using it right now; Any future issues are likely to be logic issues that should be handled anyhow imo.

I'm not fully confident if the last 3 commits are something that are enjoyable to others, but I do generally like being more explicit about when we do or don't need a async runtime.


piegamesde:
I can get behind most of the changes, but I think the syncification commit is a move in the wrong direction. The reason why `freeze` and `unfreeze` are async but do not make use of it yet is that I haven't come around to asyncify the IO lock file reading and writing, so making them sync again is backwards IMO. Especially `get_path` should not be sync IMO.

For context, I use npins and libnpins on lockfile with thousands on pins, so operations that may be performed concurrently at scale on all pins should preferably be fully async if possible

Coca162:
I'm curious to see what async IO on the lock file would look like since `tokio::fs::File` is ultimately a wrapper over the synchronous `std::fs::File`. `freeze`/`unfreeze` are also not very beneficial to be concurrent (since well, setting a single value for multiple pins is already very fast at our scale) and `get-path` only works on a single pin (and if it were to work on multiple it'd believe it'd be better to modify `nix_eval_pin` to eval something that returns multiple pins instead of 1).

I'll still get rid of the changes since sync `nix_eval_pin` is a bit questionable for the libnpins API.

piegamesde:
> tokio::fs::File is ultimately a wrapper over the synchronous std::fs::File.

That however doesn't mean that tokio `File` is an async wrapper around a sync API, at least not in the way that e.g. the `blocking` crate wraps sync code. Instead, it uses some unsafe code to make the file descriptor itself async, changing the semantics of the syscalls the `File` functions emit. Where the calls normally are sync in the sense of blocking, they are now sync in the sense that they always return immediately without ever blocking: Should a syscall block, it would now emit EWOULDBLOCK instead. The tokio wrapper then handles those cases by hooking the fd into the async runtime, to get notified when progress can be made on the fd without blocking again.

(At least that's my rough understanding from back when I did a deep dive on these things a couple of years ago. Could be that things have changed since then with epoll etc., but I don't think so)

piegamesde (approved):

Coca162:
As for as I understand it this isn't true for tokio since nowhere I can find does it make use of non blocking when creating it (tokio::fs::File::open just calls as asyncified version of std open).

For reading from a tokio File the code also seems to be blocking code as [`<tokio::fs::File as AsyncRead>::poll_read`](https://docs.rs/tokio/latest/src/tokio/fs/file.rs.html#596-661) calls the synchronous [`Buf::read_from`](https://docs.rs/tokio/latest/src/tokio/io/blocking.rs.html#241-268) in a `tokio::task::spawn_blocking` to not have blocking on the runtime.

edit: Wait, I got myself very confused now, that is not the same code snippet I saw as in my IDE when hunting this down. I'll try to figure out what I'm looking at.

Coca162:
Ah, there is a little more indirection in new releases of tokio due to the unstable io_uring support but in [`Inner::poll_read_inner`](https://docs.rs/tokio/latest/src/tokio/fs/file.rs.html#953-992) you can see the `Self::spawn_blocking_read` at the end that does the same thing as before.

piegamesde:
Oh, fascinating. Then I might be misremembering from another async runtime that did it like I described, possibly async_std (as I was using that a lot in that time).

Edit: I just looked at the source code, async_std uses spawn_blocking as well. Maybe I am misremembering, or maybe the implementation got changed because it turned out to be unsound or something?

Either way I don't care too deeply about the lock file reading and writing being async, because even when using blocking code in a future for that it should never block for too long. The main benefactor of async in npins is all the fetching and downloading code, and the subprocess spawning (though we'd ideally replace that with our own code eventually)

Coca162:
Random trivia time is over for async file reading, I understand its not really important here but its fun to look at :D

PR #234, opened 2026-05-17
Release 0.4.1

Coca162:
Since we have had a lot of major changes come in, this release will be based on a rebased branch (https://github.com/andir/npins/tree/v0.4.1) with these commits to master being to keep it in sync. [Githubs diff](https://github.com/Coca162/npins/compare/0.4.1-prep...andir:npins:v0.4.1) is annoyingly inaccurate but I have gone over it locally to verify the diff is only the changes I intentionally kept out.

Github release draft (not public): https://github.com/andir/npins/releases/tag/untagged-85e67251cc4630ed3ad3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,23 @@
   Existing locked tarball pins should remain functional just the same as before, no action required.
 - Added `npins add url` for pinning plain URLs (files that should not be unpacked as tarballs). Same semantics as `tarball` pins. (https://github.com/andir/npins/pull/223)
 - `npins add channel` now takes `--artifact` to pin something other than `nixexprs.tar.xz`, for example `latest-nixos-graphical-x86_64-linux.iso`. (https://github.com/andir/npins/pull/223)
-- `npins show` now accepts a list of pin entries to show instead of always showing the complete list (https://github.com/andir/npins/pull/190)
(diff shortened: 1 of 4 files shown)


Coca162:
https://github.com/andir/npins/commit/d134912b943a31bf80036c0ff305e1836afec142 and https://github.com/andir/npins/commit/cb687c3cd6851d5da2b9e99896706e1b66a630b2 editing `default.nix` (especially with a actual change in nix code) isn't particularly nice so I'd consider doing a version bump for those but then it'd kind of be a breaking change? The only change in nix code is for a bug fix too so I guess for users who are affected and want to use this release and not 0.5.0 they can just recreate `default.nix`.

piegamesde:
The changes to the `default.nix` can be simply ignored IMO (as in, included in the patch release as-is). People can use `npins upgrade` to get them regardless of the format version bump, but also it won't matter for most people.

Not sure what exactly your plan for the release branch is, but just go ahead with what you think makes sense

Coca162:
> Not sure what exactly your plan for the release branch is, but just go ahead with what you think makes sense

Just making a tag and then deleting it :D

Not sure there is anything else that makes sense to be done, it can be recreated it for whatever reason we need a `0.4.2`.

Issue #141, opened 2025-05-12
RFC: Patching pins


piegamesde:
Being able to patch inputs is a feature that regularly comes up now and then, so let's track it here. The difficult part is not implementing it, but designing it in a way that is actually pleasant to use. Therefore, I ask everyone who is interested in that functionality: What is your use case? What workflow would you want?

Tom-Hubrecht:
Usecases:
- I often need functionnality that is not (yet) included in the upstream sources that I pin (e.g. unmerged PRs or local customization) and patching the source is the cleanest way to do it (

Current workflow: 
- I have a nix file where I define what patches are applied to what inputs: https://git.dgnum.eu/DGNum/infrastructure/src/branch/main/patches/default.nix
- The available sources are patched here https://git.dgnum.eu/DGNum/infrastructure/src/branch/main/bootstrap.nix

However, this needs to perform IFD as a base version of nixpkgs is needed to use `applyPatches`

leana8959:
I was trying to patch a fetched source and found [pkgs.applyPatches](https://noogle.dev/f/pkgs/applyPatches). Just wanted to share considering that it's not very well-known imo in case it helps :)

Update: my bad, didn't read the previous message thoroughly.

Janik-Haag:
I have two use cases for this, one is unmerged PRs, which Tom already covered, and the other one is vendored patches.

For sources originating from github/forgejo/gitlab/etc... it would be pretty neat if npins had native support for "patches"/"cls"/"prs"/"mrs"/..., this could probably be implemented as a list of PRs in the sources.json for the given source which get updated in lock step with the matching pin. (Tangentially related #219)
E.g. you would probably do something like `npins patch nixpkgs --pr 12345`. And then the sources.json might look something like:
```
    "nixpkgs": {
      "type": "Git",
      "repository": {
        "type": "GitHub",
        "owner": "nixos",
        "repo": "nixpkgs",
        "prs": [
          {
            number: 12345,
            ...
          }
        ]
      },
      ...
    },
```

I have some machinery for a few years now that goes and reads npins/patches/${sources}/$some.patch
And then automatically applies them using pkgs.applyPatches to a given source.
```
npins/patches
├── home-manager
├── lix
├── mmp
├── nixos-mailserver
├── nixpkgs
├── ...
└── terranix
```

I have a bootstrap.nix which does all the importing, applies patches, overlays, etc..., prepare imports and then exposes those to different parts of the repo.

[...]

PR #238, opened 2026-05-30
add patches attribute for pins

Janik-Haag:
Some glue code to make patching pins more convenient, motivated by  #141.

I'm not too happy with passing pkgs and patches to every pin individually, but that seems like the most obvious path forward, since the api already exists.

Also, not sure if we want to couple patches and nixpkgs fetchers.
If we want to decouple them, but keep the same api, I'd probably add an attribute like `nixpkgsFetcher ? true` to `mkSource`. This would keep the existing api compatible but would also allow the user to toggle the behavior if they wish to. I can imagine this being useful for the nixpkgs pin, so you can patch it without having to fetch it twice.

The test is a bit shit, before merging it should probably be improved. Currently, if the patch can't be applied it still succeeds and is completely silent about the failure in the logs (bash is horrible).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Pins now optionally take a patches attribute and apply them using pkgs.applyPatches
+
 ## 0.5.0
 
 - **[Breaking] Support for the "Lockable HTTP Tarball Protocol" has been removed.**
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Simple and convenient dependency pinning for Nix
   - Unlike tracking a channel from its git branch, this gives you access to the `programs.sqlite` database
   - Can also track Nix channel artifacts like live isos
 - Track PyPi packages
+- Patch pins with pkgs.applyPatches
 
 ## Getting Started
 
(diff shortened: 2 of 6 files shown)


piegamesde:
@Coca162 @andir what do you think about this? I'm a bit concerned that the API is merely a thin wrapper over `applyPatches`, and thus does not provide much for users while having maintenance cost. At the same time, I see that any more fully featured implementation of patching would be a *lot* more effort, requiring lots of new CLI commands etc.

One additional complication is that unfortunately, patching relies on having a Nixpkgs, and this will get weird trying to get our `get-path` to work with that …

One very minimal variant of this PR would be to only document how to patch pins with `applyPatches` in the README, without adding special support in `default.nix`

Coca162:
Another concern with having this built into npins' glue code is that `applyPatches` is unnecessary for anything that goes through `mkDerivation` and otherwise IFD for nix sources.

andir:
I don't know. I see the appeal but how many people need this, and how often is this actually what you want? Any large project would probably not want to do it this way as that is likely super expensive IFD wise and patching sources for packaging can be done with nixpkgs machinery much nicer.

andir:
If someone needs to routine add like patch series to a repository they should probably write a script that does just that and commit the generated files.

RossSmyth:
If you mean `pkgs.applyPatches` by
> patching sources for packaging can be done with nixpkgs machinery much 

That is true in some cases, but it's also pretty easy to get infrec if you try to patch things as well. Specifically when you have module imports.

andir:
Sorry. I still think this is out of scope.

PR #239, opened 2026-05-31
Migrate to bpaf command line parsing and completions

Coca162:
Lots of cleanup needed, I'm also not happy with how there is some bugginess and differences in the completions compared to with the native fish completions. There is also no `help` subcommands currently and the help formatting is different in ways I minorly dislike. All flags are global as well, though this is not indicated in help. Hopefully some of these can be dealt with upstream.

Otherwise this was library was pretty fun to work with even including the relatively cursed way to share the lockfile to completions. Combinatorics are much nicer then what clap offers imo.


Coca162:
Changing the nix tests so that `npins --lock-file sources.json -d npins` isn't accepted anymore, since bpaf can easily handle directory having a default and being mutually exclusive with another flag unlike clap

pacak:
o/

Upstream here :)

> I'm also not happy with how there is some bugginess and differences in the completions compared to with the native fish completions.

Not happy with that either. Originally I had it generating native fish completions, turns out fish caches them between runs instead of asking for a new set every time leading to wrong completions being offered. There was a ticket open about this specific issue and it was closed with "well, nobody does dynamic completions like that so current behavior is fine". Duh. I wonder why nobody does it....

But that was in C++ days, maybe it behaves differently now. Will revisit at some point now that I have more robust testing system for completions in place.

> There is also no help subcommands currently

You can make one yourself.

To support stuff like `npins help add tarball` you can add a variant with command `help` and consume everything in the body with `any` or something, then do `OptionParser::run_inner` on the input with `--help` added at the back. Or instead of adding `help` to your commands make a wrapper like `enum HelpOrCmd { Help(Vec<OsString>`), Command(Command)`.

> and the help formatting is different in ways I minorly dislike.

Kind of hard to do something about it at the moment. Slowly reworking some bits to make it easier to tweak it, but no ETA promises.

> All flags are global as well, though this is not indicated in help.

Same as with changing help formatting. There's something that works, but I'm planning to revisit most of the documentation. No ETA.

> Hopefully some of these can be dealt with upstream.

Some :)

> Otherwise this was library was pretty fun to work 

:heart:

> with even including the relatively cursed way to share the lockfile to completions. 

Should become less cursed in the next major release (still no ETA)

> Combinatorics are much nicer then what clap offers imo.

Lax Monoidal Functors FWIW.

src/opts.rs
```diff
@@ -14,7 +24,20 @@ pub enum UpdateStrategy {
...
+impl UpdateStrategy {
+    pub fn parser() -> impl Parser<Self> {
+        let full = long("full").short('f').help("Re-fetch hashes even if the version hasn't changed.\nUseful to make sure the derivations are in the Nix store.").req_flag(Self::Full);
+        let partial = long("partial")
+            .short('p')
+            .help("Don't update versions, only re-fetch hashes")
+            .req_flag(Self::HashesOnly);
+        let normal = pure(Self::Normal);
+
+        construct!([full, partial, normal])
```
pacak: This should give the same parsing behavior, but slightly better error messages if something goes wrong.

```suggestion
        construct!([full, partial]).fallback(Self::Normal)
```

src/opts.rs
```diff
@@ -24,100 +47,166 @@ pub struct ChannelAddOpts {
...
 pub struct GitHubAddOpts {
-    #[arg(value_hint = ValueHint::Other)]
+    // #[arg(value_hint = ValueHint::Other)]
     pub owner: String,
-    #[arg(value_hint = ValueHint::Other)]
+    // #[arg(value_hint = ValueHint::Other)]
     pub repository: String,
 
-    #[command(flatten)]
+    // #[command(flatten)]
```
pacak: if you want to use derive macro - it would be `#[bpaf(external)]`

src/opts.rs
```diff
@@ -132,219 +221,422 @@ pub enum GitForgeOpts {
...
+                    "none" => Self::None,
+                    "auto" => Self::Auto,
+                    "gitlab" => Self::Gitlab,
+                    "github" => Self::Github,
+                    "forgejo" => Self::Forgejo,
+                    x => return Err(format!("invalid value '{x}' for forge")),
+                })
+            })
+            .fallback(GitForgeOpts::Auto)
+            .format_fallback(|_, f| f.write_str("auto"))
```
pacak: `.display_fallback()` will probably do the same if for `GitForgeOpts::Auto` `std::fmt::Display` says "auto".

src/opts.rs
```diff
@@ -132,219 +221,422 @@ pub enum GitForgeOpts {
...
+        })
+        .map(Self::Channel)
+        .to_options()
+        .descr("Track a Nix channel")
+        .command("channel");
+
+        let github = construct!(GitHubAddOpts {
+            more(GenericGitAddOpts::parser()),
+            owner(positional("OWNER")),
+            repository(positional("REPOSITORY"))
```
pacak: This syntax is going away in the next big release (no ETA), I think it's confusing after all.

src/opts.rs
```diff
@@ -132,219 +221,422 @@ pub enum GitForgeOpts {
...
+        .map(Self::Git)
+        .to_options()
+        .descr("Track a git repository")
+        .command("git");
+
+        let pypi = construct!(PyPiAddOpts {
+            at(long("at").argument("VERSION").help("Use a specific release instead of the latest.").optional()),
+            version_upper_bound(long("upper-bound").argument("VERSION").help(r#"Bound the version resolution. For example, setting this to "2" will restrict updates to 1.X versions."#).optional()),
+            package_name(positional("PACKAGE").help("Name of the package at PyPi.org"))
+        })
```
pacak: And that's the example why I don't like this syntax :)

Compact, but no rust-analyzer, no rust-fmt, etc. Having a few variables defined outside and referring them from the macro is cleaner

```rust
let at = long("at")... 
...
let pypi = construct!(PyPiAddOpts { at, version_upper_bound, package_name });`
```

Coca162: The temptation of trying to make it concise got to me! I definitely agree with your assessment that its a bit scuffed though, I have similar reasons as to why I don't like derive macros as well.

src/opts.rs
Coca162: Yep yep, just went for the quickest route here but might as well

src/opts.rs
Coca162: Thanks!

PR #243, opened 2026-07-06
optimize `update` for container

felixscheinost:
Currently when doing `npins add` or `npins update` for container, `nix-prefetch-docker` is always called twice.

First with the tag `nix-prefetch-docker <image> <tag>` then with the returned digest (`nix-prefetch-docker <image> <digest>`).

But the first response already contains the hash.

The practical reason for this beahviour is that `Updateable` trait has two operations, that need to be chained to do an update: `update` and `fetch`.

This is also reflected in then types:

update: ContainerVersion -> ContainerVersion

fetch: ContainerVersion -> ContainerHash

So in order to skip the second fetch, we can either do a major change to he codebase. Maybe have a trait that is related to `Updateable` but that signals that `update` and `fetch` are the same. I didn’t want to do that.

A more pragmatic way was to find a way to sneak the missing data into `ContainerVersion` so that `fetch` can populate `ContainerHash` solely from `ContainerVersion`.

My solution was to add a field to `ContainerVersion` which isn’t serialized. When `ContainerVersion` is a freshly fetched version from `update`, then `fetch` can simply return that hash.

---

Logs from `master`

```
$ cargo run -- add container amazon/aws-otel-collector v0.47.0 --arch amd64 --verbose
[INFO ] Adding 'amazon/aws-otel-collector' …
[DEBUG] Executing: nix-prefetch-docker amazon/aws-otel-collector v0.47.0 --json --quiet --arch amd64
[DEBUG] nix-prefetch-git output: {
      "imageName": "amazon/aws-otel-collector",
      "imageDigest": "sha256:40a7eb9bfa58871d5d6dbdfcb146bc9d7eb4f149b42bd653809cd5b980af0cc4",
      "hash": "sha256-HEJ1p67aErWbcUaIQUDn7n5Ed1Rl3Vo3+ysCXHUtB6I=",
      "finalImageName": "amazon/aws-otel-collector",
      "finalImageTag": "v0.47.0"
    }

[DEBUG] Executing: nix-prefetch-docker amazon/aws-otel-collector v0.47.0 --json --quiet --image-digest sha256:40a7eb9bfa58871d5d6dbdfcb146bc9d7eb4f149b42bd653809cd5b980af0cc4 --arch amd64
[DEBUG] nix-prefetch-git output: {
      "imageName": "amazon/aws-otel-collector",
      "imageDigest": "sha256:40a7eb9bfa58871d5d6dbdfcb146bc9d7eb4f149b42bd653809cd5b980af0cc4",
      "hash": "sha256-HEJ1p67aErWbcUaIQUDn7n5Ed1Rl3Vo3+ysCXHUtB6I=",
      "finalImageName": "amazon/aws-otel-collector",
      "finalImageTag": "v0.47.0"
    }

    image_name: amazon/aws-otel-collector
    image_tag: v0.47.0
    arch: amd64
    image_digest: sha256:40a7eb9bfa58871d5d6dbdfcb146bc9d7eb4f149b42bd653809cd5b980af0cc4
    hash: sha256-HEJ1p67aErWbcUaIQUDn7n5Ed1Rl3Vo3+ysCXHUtB6I=
    frozen: false
```

---

Logs after this patch:

```
 $ cargo run -- add container amazon/aws-otel-collector v0.47.0 --arch amd64 --verbose
[INFO ] Adding 'amazon/aws-otel-collector' …
[DEBUG] Executing: nix-prefetch-docker amazon/aws-otel-collector v0.47.0 --json --quiet --arch amd64
[DEBUG] nix-prefetch-git output: {
      "imageName": "amazon/aws-otel-collector",
      "imageDigest": "sha256:40a7eb9bfa58871d5d6dbdfcb146bc9d7eb4f149b42bd653809cd5b980af0cc4",
      "hash": "sha256-HEJ1p67aErWbcUaIQUDn7n5Ed1Rl3Vo3+ysCXHUtB6I=",
      "finalImageName": "amazon/aws-otel-collector",
      "finalImageTag": "v0.47.0"
    }

    image_name: amazon/aws-otel-collector
    image_tag: v0.47.0
    arch: amd64
    image_digest: sha256:40a7eb9bfa58871d5d6dbdfcb146bc9d7eb4f149b42bd653809cd5b980af0cc4
    hash: sha256-HEJ1p67aErWbcUaIQUDn7n5Ed1Rl3Vo3+ysCXHUtB6I=
    frozen: false
```


piegamesde:
The correct solution for this would be to have `nix-prefetch-docker` give us the hash of the current version without having to download it first, or any other tool which implements that functionality. I don't know enough about Docker fetching and Docker registries to be able to tell how feasible that is.

piegamesde:
Okay it seems to be quite simple:

Our `update` function only takes the `image_digest` field from the tool: https://github.com/andir/npins/blob/master/libnpins/src/pins/container.rs#L59-L61, which is calculated with a simple `skopeo inspect` command call in `nix-prefetch-docker`: https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/docker/nix-prefetch-docker#L34-L41. npins could inline that call and thus eliminate the double-download without having to resort to hacks

felixscheinost:
I don't think it is that simple.

The download isn't about the `image_digest`, it is about the Nix sha256 hash.

I think for that we have to download the image, as the container registry doesn't return that hash and we can't compute it from the metadata (I believe?).

felixscheinost:
Ahh, I see now (in your commit in #251) what you meant 🙈 

The thing is that yeah while we DO need to download to get the Nix hash, `Pin#update` doesn't need to fetch the digest yet.

Your solution using skopeo directly looks sensible. Do we need to ensure that `skopeo` is in the PATH now? `skopeo` needs to be added [here](https://github.com/NixOS/nixpkgs/blob/8623c4c20aa4ca2f5fb81510d2944066c3fb0d96/pkgs/by-name/np/npins/package.nix#L14)?

piegamesde:
> The thing is that yeah while we DO need to download to get the Nix hash, Pin#update doesn't need to fetch the digest yet.

Exactly, `update` only needs to find out what the latest version is in the first place. In rough Nix terms, this is input-addressed vs output-addressed (update pins which input to pin, while fetch produces the fixed-output hash)

> Do we need to ensure that skopeo is in the PATH now? skopeo needs to be added [here](https://github.com/NixOS/nixpkgs/blob/8623c4c20aa4ca2f5fb81510d2944066c3fb0d96/pkgs/by-name/np/npins/package.nix#L14)?

Yes and yes

PR #251, opened 2026-07-27
Rewrite test suite, import Lon, and more

piegamesde:
Oh boy, it's been a slightly manic week and weekend of hyperfocusing on npins, but here we are. (Expect this to be my yearly contribution to this project in advance for the rest of the year :D) Git history isn't the best, I can split this up into more commits if you want to, otherwise just review this by-commit Gerrit style.

We now have proper snapshot testing for the CLI! And the git tests run against a real forgejo instance!

CC @felixscheinost for #243, @CommentatorForAll for the Python code, @Tom-Hubrecht for the lon importer (I don't really grok the format, I've glanced at the cross-stream source code and bashed rocks together until the tests passed, so be warned)


tests/import.py
```diff
@@ -0,0 +1,207 @@
...
+    client.fail("npins import-flake --name dependency flake/flake.lock")
+
+with subtest("import lon"):
+    clean_workdir()
+    client.succeed("lon init")
+    client.succeed("lon add github owner/dependency test-branch")
+    client.succeed("lon add git tagged-repo http://localhost/tagged-repo.git test-branch")
+    client.succeed("lon add git submoduled-repo http://localhost/submoduled-repo.git main --submodules")
+    client.succeed("lon add git frozen-repo http://localhost/untagged-repo.git foo --frozen")
+    # Lon has no CLI command for adding tarball sources (they only come from importing from other tools)
```
Tom-Hubrecht, line 127: There is, `lon add tarball <NAME> <URL>`. If you don't have it, it would be because your version is outdated

piegamesde: Ah lol, good thing I added the test anyways then (though could have saved myself the effort of faking the pins) ^^

piegamesde: Okay the version was released just yesterday, so that's literally before I wrote this code ^^

Tom-Hubrecht: Oh, to be fair, I use the main branch and not a release ^^'

Coca162:
Trying out the tests myself and I am getting a hash mismatch on testTarball, it might be better to vendor it instead of generating it in nix if its because the derivation isn't reproducible.

```
error: builder for '/nix/store/1p7byg1p2nyvv9f8q6hs9x4421vs6nr6-npins-snapshot-check-cli.drv' failed with exit code 1;
       last 14 log lines:
       > diff --unified --recursive /nix/store/1slwfjn1qxs6qw62f8314r58k84vzndv-snapshots_cli/get_path_bad_hash.exp /nix/store/phd5vm9fh2fljb80gnzyy80zy6bxns90-vm-test-run-npins-cli-test/snapshots/get_path_bad_hash.exp
       > --- /nix/store/1slwfjn1qxs6qw62f8314r58k84vzndv-snapshots_cli/get_path_bad_hash.exp        1970-01-01 00:00:01.000000000 +0000
       > +++ /nix/store/phd5vm9fh2fljb80gnzyy80zy6bxns90-vm-test-run-npins-cli-test/snapshots/get_path_bad_hash.exp 1970-01-01 00:00:01.000000000 +0000
       > @@ -57,7 +57,7 @@
       >
       >         error: hash mismatch in file downloaded from 'http://localhost/testTarball':
       >           specified: sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
       > -         got:       sha256-uu7ApyKPllAihGSgrXQF9y/Gx8Vut9cpLn+4T82n+EE=
       > +         got:       sha256-I1SJhtKxecbedOw0hDnNcW3ydyaQQowpnZ0v2vy0RoA=
       >  Error: Could not evaluate pin
       >
       >  Caused by:
       > error: snapshot mismatch for test 'cli'
       > Run `just update-snapshots` to accept the changes
```

Edit: GNU tar does have a whole section about reproducibility if we wanted to keep them in nix https://www.gnu.org/software/tar/manual/html_node/Reproducibility.html

libnpins/src/import/niv.rs
```diff
@@ -0,0 +1,70 @@
...
+    branch: String,
+    /* Either owner is None repo is the git URL, or owner is Some and then it's a GitHub owner/repo style thing */
+    repo: String,
+    owner: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NivUrlPin {
+    #[serde(rename = "type")]
+    type_: String, // Should be only "tarball" or "file", but who knows
```
Coca162: nit: `r#type` is a little less clunky to me here (serde will treat it as `type`)

piegamesde:
@Coca162 I wrote the tests with the goal of being hash-agnostic, hence using things like `ls-remote` at runtime. (This is a lesson learned the hard way back when we debugged some npins CI failures that traced down to regressions in how CppNix hashes single-file folders.) However, I failed to think of the snapshot tests subverting that. My proposed solution would be to instead apply some post-processing to the snapshot output that strips hashes to make things more deterministic again. This is also what Lix does for things like paths. What do you think?

Coca162:
I think I lean towards holding off on doing post-processing to see how necessary I'd be when we update the tests. This issue is rather small to fix and if stripping is to be done wouldn't it also have to include things like commits and URLs?

tests/import.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # Import only a single entry into a second lockfile
+    client.succeed("npins --lock-file sources2.json init --bare")
+    client.succeed("npins --lock-file sources2.json import-niv --name tagged-repo")
+    assert set(dump_pins("sources2.json").keys()) == {"tagged-repo"}
+    client.fail("npins import-niv --name tagged-repo")
+    client.fail("npins import-niv --name does-not-exist")
+
+with subtest("import flake"):
+    clean_workdir()
```
andir: That seems like a code smell. If we have to manually cleanup the working directory. The test harness should always start clean and the initial state should be defined.

test.nix
```diff
@@ -1,917 +1,81 @@
...
+      openssl
+      gitMinimal
+      dockerTools
+      npins
+      ;
+  };
+
+  # Note: We must use VMs here because Nixpkgs doesn't support Lix currently,
+  # See <https://git.lix.systems/lix-project/lix/issues/1243>
+  mkTest =
```
andir, line 27: So we always require a VM for all tests because Lix is broken? Not sure this is worth the trouble. `npins` shouldn't be partial in the "fight" for the "correct" Nix implementation.. We could just not use Lix here and avoid fighting bugs.

piegamesde: Lix isn't "broken", the container tests require an experimental feature, for which the implementations in CppNix and Lix haven't converged yet. So "just not use Lix here" would require starting a second non-Lix daemon, which isn't as easy as "just not use Lix" in the dev shell

tests/import.py
piegamesde: Agreed it is. Unfortunately things are made somewhat ugly because the test harness doesn't support properly setting PWD. Still, I could write our own `subtest` function which takes care of cleaning up state

tests/import.py
piegamesde: I've decided to not bother with this until lon actually hits 1.0 in Nixpkgs

test.nix
piegamesde: I'll see if I can come up with something that allows both kinds of tests without too much effort, because I do agree that currently the new CI times are indeed painful

piegamesde:
@Coca162 re the hashes, I'd like to keep all test artifacts declaratively generated instead of checked in. This may sound petty for a simple tarball, but for larger things like the container test image I'd really like to avoid it. (This may be paranoid, but remember how smuggling malware through checked-in binary test artifacts was used as part of the xz attack)

I've now simply tried to make the tarball more deterministic, let's hope that this is enough. If it doesn't, we can still escalate into two directions:

1. Make test artifacts output-addressed to have hash mismatches blow up earlier and louder
2. Adapt the test suite to "know" what the hashes will be and do a substitution on the snapshot output (this is better than a generic "hash-detector" in the sense that it is much more specific and thus will have no false positives)

Coca162:
I do agree with the reasoning to not vendor test artifacts and the substitution route sounds good if it becomes necessary later on

piegamesde:
This took me a lot more debugging and iterations than expected, but things should be working now. Indeed I did not manage to make the tarball deterministic, so I'm sanitizing that specific hash in the snapshots output and it's working now.

test.nix
piegamesde: Tests are now using containers by default, with an override for VM tests instead, and lix-specific entry points for devs

tests/import.py
piegamesde: We now have `npins_subtest` and our own `succeed` and `fail` helper functions which take care of everything

piegamesde:
I'd like for this to not get lost and rot, so I'd consider merging if y'all don't mind

Coca162 (approved):
I was meaning to try the nixcpp testing locally as well but then I got caught up with other things. Issues with that should show up in CI anyways so its fine.

All the diifs since last time I tested it look good though.

Issue #254, opened 2026-08-06
update: Only logs changed sources when output goes to a terminal


antifuchs:
When running `npins update` in cron, I used to get an overview of the pins that actually changed in the operation; that is no longer the case, since I'm redirecting npins' output to a file so that it can be part of the commit message. E.g.,

```console
$ npins update 2>&1 | cat
[INFO ] Update successful.
```

...in an update that changed 5 pins. When running on the console, this information is present (just the one pin for example's sake):

```console
[wrtag] Changes:
-    revision: 231eb9cca84a373a168996b2568a6f52be01d43f
+    revision: f409ecc5c2fa0e4bb53679c9cd0787aab7bda743
-    url: https://github.com/sentriz/wrtag/archive/231eb9cca84a373a168996b2568a6f52be01d43f.tar.gz
+    url: https://github.com/sentriz/wrtag/archive/f409ecc5c2fa0e4bb53679c9cd0787aab7bda743.tar.gz
-    hash: sha256-M+HzGCcTWObLwE5W5cSDUTiIy8eCjw+AsIjOIeL9vrQ=
+    hash: sha256-F8mWArIeiL3aV4j1aDam8ROeVHOx3+4UaANHDfElAEQ=
```

When using `--verbose`, npins does log information about changed pins (since they're the only ones that do change), but the information listing the the old revision/new revision is not present there either.

Should the thing listing the changes be updated to log at INFO level, maybe?

antifuchs:
Looking at the source a little more closely, it looks like `write_diff` is invoked only on the code path where stderr is a terminal, via `animation.on_pin_finish`. Might need some restructuring for the non-terminal case.

Coca162:
Yeah the writing out to stderr for the diff is cut off early alongside all the animation if we detect its not a terminal. Putting `result_writer(stderr)` in the non-terminal case should be enough to fix it.

calliecameron:
Any chance of backporting this fix into the version in stable nixpkgs? I'm also using npins in a cronjob and this would make things easier.

Coca162:
When a new version comes out with the fix I'll look into a backported one as well, shouldn't be too bad to cherry pick to it.

PR #258, opened 2026-08-30
Release 0.5.1

Coca162:
Think we have enough things to make a new release wanted. 

v0.4.2 will also be released for stable in [this branch](https://github.com/andir/npins/tree/v0.4) since it was asked for in https://github.com/andir/npins/issues/254#issuecomment-5245057571


CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
 # Changelog
 
-## Unreleased
```
piegamesde, line 3: Nit: I like to keep the `## Unreleased` line as the invitation for the next person to write something in there, so that the ritual doesn't get lost

